PR #1, opened 2016-09-17
Detailed SSL warnings for tls_connect and tls_accept

lgrahl:
It took me a lot of time to find out why my calls to `tls_connect` and `tls_accept` failed. Eventually I found out that my peers have used incompatible TLS versions and at some point could not negotiate a cipher suite. These changes print out the additional information OpenSSL provides when such events occur which makes it easier to detect the cause of the problem.

### DIFF
--- a/src/tls/openssl/tls_udp.c
+++ b/src/tls/openssl/tls_udp.c
@@ -242,29 +242,29 @@ static int tls_connect(struct tls_conn *tc)
 	r = SSL_connect(tc->ssl);
 	if (r <= 0) {
 		const int ssl_err = SSL_get_error(tc->ssl, r);
-        int err = 0;
+		int err = 0;
 
 		switch (ssl_err) {
 
 		case SSL_ERROR_WANT_READ:
 			break;
-        
-        case SSL_ERROR_SYSCALL:
-        case SSL_ERROR_SSL:
-            DEBUG_WARNING("connect error: ");
-            ERR_print_errors_fp(stderr);
-            err = EPROTO;
-            break;
+		
+		case SSL_ERROR_SYSCALL:
+		case SSL_ERROR_SSL:
+			DEBUG_WARNING("connect error: ");
+			ERR_print_errors_fp(stderr);
+			err = EPROTO;
+			break;
 
 		default:
 			DEBUG_WARNING("connect error: %i\n", ssl_err);
 			err = EPROTO;
 		}
-        
-        ERR_clear_error();
-        
-        if (err)
-            return err;
+		
+		ERR_clear_error();
+		
+		if (err)
+			return err;
 	}
 
 	check_timer(tc);
@@ -282,19 +282,19 @@ static int tls_accept(struct tls_conn *tc)
 	r = SSL_accept(tc->ssl);
 	if (r <= 0) {
 		const int ssl_err = SSL_get_error(tc->ssl, r);
-        int err = 0;
+		int err = 0;
 
 		switch (ssl_err) {
 
 		case SSL_ERROR_WANT_READ:
 			break;
-        
-        case SSL_ERROR_SYSCALL:
-        case SSL_ERROR_SSL:
-            DEBUG_WARNING("accept error: ");
-            ERR_print_errors_fp(stderr);
-            err = EPROTO;
-            break;
+		
+		case SSL_ERROR_SYSCALL:
+		case SSL_ERROR_SSL:
+			DEBUG_WARNING("accept error: ");
+			ERR_print_errors_fp(stderr);
+			err = EPROTO;
+			break;
 
 		default:
 			DEBUG_WARNING("accept error: %i\n", ssl_err);
@@ -302,9 +302,9 @@ static int tls_accept(struct tls_conn *tc)
 		}
 
 		ERR_clear_error();
-        
-        if (err)
-            return err;
+		
+		if (err)
+			return err;
 	}
 
 	check_timer(tc);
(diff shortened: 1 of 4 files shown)


alfredh:
Hi Lennart,

thanks for your patch, I think this is a useful feature to add. I have also done a lot
of DTLS debugging and have had to add this type of extra debugging to find why
a DTLS handshake failed.

could you please fix all the ccheck errors:

```
./src/tls/openssl/tls_udp.c:245: starts with 8 spaces, use tab instead
./src/tls/openssl/tls_udp.c:251: has trailing space(s)
./src/tls/openssl/tls_udp.c:252: starts with 8 spaces, use tab instead
./src/tls/openssl/tls_udp.c:253: starts with 8 spaces, use tab instead
./src/tls/openssl/tls_udp.c:254: starts with 12 spaces, use tab instead
./src/tls/openssl/tls_udp.c:255: starts with 12 spaces, use tab instead
./src/tls/openssl/tls_udp.c:256: starts with 12 spaces, use tab instead
./src/tls/openssl/tls_udp.c:257: starts with 12 spaces, use tab instead
./src/tls/openssl/tls_udp.c:263: has trailing space(s)
./src/tls/openssl/tls_udp.c:264: starts with 8 spaces, use tab instead
./src/tls/openssl/tls_udp.c:265: has trailing space(s)
./src/tls/openssl/tls_udp.c:266: starts with 8 spaces, use tab instead
./src/tls/openssl/tls_udp.c:267: starts with 12 spaces, use tab instead
./src/tls/openssl/tls_udp.c:285: starts with 8 spaces, use tab instead
./src/tls/openssl/tls_udp.c:291: has trailing space(s)
./src/tls/openssl/tls_udp.c:292: starts with 8 spaces, use tab instead
./src/tls/openssl/tls_udp.c:293: starts with 8 spaces, use tab instead
./src/tls/openssl/tls_udp.c:294: starts with 12 spaces, use tab instead
./src/tls/openssl/tls_udp.c:295: starts with 12 spaces, use tab instead
./src/tls/openssl/tls_udp.c:296: starts with 12 spaces, use tab instead
./src/tls/openssl/tls_udp.c:297: starts with 12 spaces, use tab instead
./src/tls/openssl/tls_udp.c:305: has trailing space(s)
./src/tls/openssl/tls_udp.c:306: starts with 8 spaces, use tab instead
./src/tls/openssl/tls_udp.c:307: starts with 12 spaces, use tab instead
```

src/tls/openssl/tls_udp.c
```diff
@@ -242,18 +242,29 @@ static int tls_connect(struct tls_conn *tc)
...
 
 		switch (ssl_err) {
 
 		case SSL_ERROR_WANT_READ:
 			break;
+		
+		case SSL_ERROR_SYSCALL:
+		case SSL_ERROR_SSL:
+			DEBUG_WARNING("connect error: ");
+			ERR_print_errors_fp(stderr);
```
alfredh: one thing to consider, should the openssl errors be printed to STDERR or to the
DEBUG_xxx system? The Debug system has callback handlers so that the application
can redirect debug message to e.g. a file.

src/tls/openssl/tls_udp.c
```diff
@@ -242,18 +242,29 @@ static int tls_connect(struct tls_conn *tc)
...
+			err = EPROTO;
+			break;
 
 		default:
 			DEBUG_WARNING("connect error: %i\n", ssl_err);
-			return EPROTO;
+			err = EPROTO;
```
alfredh: please add a `break` here.

src/tls/openssl/tls_udp.c
lgrahl: I think it should be using the debug system. (To be honest, I never figured out how to use the re_*print functions properly so I'm not sure how this should be done.)

alfredh:
Hi Lennart,

the openssl api contains a couple of functions for printing the error queue (err.h):

```
void ERR_print_errors_cb(int (*cb) (const char *str, size_t len, void *u),
                         void *u);
# ifndef OPENSSL_NO_FP_API
void ERR_print_errors_fp(FILE *fp);
# endif
# ifndef OPENSSL_NO_BIO
void ERR_print_errors(BIO *bp);
# endif
```

how about using the callback function to print errors:

```
static int print_errors_cb(const char *str, size_t len, void *u)
{
	DEBUG_WARNING("%b", str, len);
	return 0;
}

{
	/* ... */
	DEBUG_WARNING("connect error: %d\n", ssl_err);
	ERR_print_errors_cb(print_errors_cb, NULL);
	/* ... */
}

```

lgrahl:
Thanks for digging into OpenSSL's code. :)

Two short questions:

1. Would it make sense to call `ERR_print_errors_cb` in the `default` section so that other errors also produce detailed warnings?
2. `tls_accept` and `tls_connect` are nearly identical. Is there anything that speaks against moving the error handling code into a separate function?

alfredh:
it would be better to call the ERR_print_errors_cb from the default section, yes ..

lgrahl:
Alright, will change that when I'm back on my PC. Could you comment on the second question as well?

alfredh:
this patch looks fine to me.

is it possible to squash all commits into a single one? it makes the history look better
and it is easier to revert in case of mistakes.

richaas:
I think we can simply replace ERR_clear_error() with an alternative function that prints and clears.
In the case of SSL_ERROR_WANT_READ, nothing is printed. In addition to print, ERR_print_errors_cb() also clears the error queue (like we do with ERR_clear_error()). I have attached a proposed patch.
[sslerr.diff.txt](https://github.com/creytiv/re/files/485131/sslerr.diff.txt)

lgrahl:
@richaas I think your solution fits even better. Do you want to open another PR with the proposed changes or should I do it?

richaas:
Feel free to just update this PR.

lgrahl:
I've replaced all `ERR_clear_error` calls with `tls_flush_error` in the openssl directory. One downside to the `tls_flush_error` function is that it uses the *tls* prefix for debug messages:

```
...
[000031664] dtls: accept error: 1
[000031664] tls: 139843465774912:error:1408A0C1:SSL routines:ssl3_get_client_hello:no shared cipher:s3_srvr.c:1417:
...
```

Give me some feedback on that and I'll happily squash the commits if you're all fine with the changes (@alfredh: AFAIK there should also be a button for you to squash and merge).

alfredh:
I have made a new PR for this:

https://github.com/creytiv/re/pull/8

PR #2, opened 2016-09-17
Use one DTLS connection for multiple UDP sockets

lgrahl:
These changes make it possible to use one DTLS connection over multiple specific UDP sockets which is required by WebRTC and ORTC. It is the best solution I could find with the fewest changes to the API.

For WebRTC/ORTC, as soon as another candidate pair has been found, the UDP socket can be added by calling `dtls_attach_udp_sock(<dtls-socket>, <new-udp-socket>, <layer>)`


alfredh:
Hi Lennart,

many thanks for this patch. also this feature is very useful to add.

I think the patch has 2 separate features:

* Use max 1 DTLS-connection for multiple peers
* Add ability to receive dtls packets on multiple UDP sockets

in my other webrtc-related project I have also implemented very similar code,
for example a new function `dtls_receive_packet` so the stack can receive packet
over any transport.

I think the 1 or N connection part is quite logical to add.

For the 2nd part (receive over multiple UDP-sockets) I think we should broaden the
scope a bit. For example, if the underlying transport is ICE then your ICE stack might
choose to change the active candidate-pair.

* ICE ==> UDP-socket
* ICE-TCP ==> TCP-connection
* TURN ==> UDP-socket with headroom
* TURN/TCP ==> TCP-connection
* TURN/TLS ==> TCP-connection
* SDP piggybacked (ref. draft-rescorla-dtls-in-sdp-00)

given these requirements I think we should seek a more generic solution
that can fit all of them :)

src/tls/openssl/tls_udp.c
```diff
@@ -739,6 +770,58 @@ int dtls_listen(struct dtls_sock **sockp, const struct sa *laddr,
...
+
+	err = udp_register_helper(&att->uh, us, layer,
+				  NULL, recv_handler, sock);
+	if (err)
+		goto out;
+
+	return 0;
+
+ out:
+	mem_deref(us);
```
alfredh: a small nit; I dont think this line is needed.

lgrahl:
Thanks for your feedback. I'll continue my work and see what I come up with once I add TCP support to my *RTC implementation. :)

lgrahl:
I'll close this for now.

lgrahl:
@alfredh If I understand you correctly, you'd still be interested in the changes if I strip out the "receiving DTLS packets on multiple UDP sockets"? In that case, I'll open a new PR for that - let me know.

lgrahl:
@alfredh FYI I have another branch in the pipeline which will make it possible to create a detached DTLS socket (meaning it does not need a UDP socket to be used). I'll open a PR once I've tested it with ICE-TLS but thought [you might be interested in the changes](https://github.com/creytiv/re/compare/master...rawrtc:dtls-socketless) beforehand. Feedback welcome.

PR #6, opened 2016-09-19
Add support for DragonFly BSD

czarkoff:
This was discussed on mailing list. I've tested it on dfly back then, and it worked. I had a report from someone who was successfully using baresip on dfly with these changes.

### DIFF
--- a/mk/re.mk
+++ b/mk/re.mk
@@ -548,9 +548,9 @@ HAVE_EPOLL   := $(shell [ -f $(SYSROOT)/include/sys/epoll.h ] || \
 			&& echo "1")
 endif
 
-HAVE_LIBRESOLV := $(shell [ -f $(SYSROOT)/include/resolv.h ] && echo "1")
-ifneq ($(HAVE_LIBRESOLV),)
-CFLAGS  += -DHAVE_LIBRESOLV
+HAVE_RESOLV_H := $(shell [ -f $(SYSROOT)/include/resolv.h ] && echo "1")
+ifneq ($(HAVE_RESOLV_H),)
+CFLAGS  += -DHAVE_RESOLV_H
 endif
 ifneq ($(HAVE_SYSLOG),)
 CFLAGS  += -DHAVE_SYSLOG
--- a/mk/re.mk
+++ b/mk/re.mk
@@ -273,6 +273,16 @@ ifeq ($(OS),netbsd)
 	AFLAGS		:= cru
 	HAVE_KQUEUE	:= 1
 endif
+ifeq ($(OS),dragonfly)
+	CFLAGS		+= -fPIC -DDRAGONFLY
+	LFLAGS		+= -fPIC
+	SH_LFLAGS	+= -shared
+	MOD_LFLAGS	+=
+	APP_LFLAGS	+= -rdynamic
+	AR		:= ar
+	AFLAGS		:= cru
+	HAVE_KQUEUE	:= 1
+endif
 ifeq ($(OS),freebsd)
 	CFLAGS		+= -fPIC -DFREEBSD
 	LFLAGS		+= -fPIC
--- a/mk/re.mk
+++ b/mk/re.mk
@@ -221,7 +221,7 @@ BIN_SUFFIX	:=
 
 ifeq ($(OS),solaris)
 	CFLAGS		+= -fPIC -DSOLARIS
-	LIBS		+= -ldl -lsocket -lnsl
+	LIBS		+= -ldl -lsocket -lnsl -lresolv
 	LFLAGS		+= -fPIC
 	SH_LFLAGS	+= -G
 	MOD_LFLAGS	+=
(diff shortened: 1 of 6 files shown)


czarkoff:
ping

czarkoff:
ping

alfredh:
this patch looks fine to me, I have tested it on Debian Testing.

richaas:
IMO we should aim to keep the LIBS += -lresolv within the HAVE_LIBRESOLV test.

czarkoff:
@richaas: that would make sense if `HAVE_LIBRESOLV` had anything to do with `libresolv`.  It tests for `resolv.h` instead, which says about the API, but not about the library.  I added a commit to resolve this ambiguity.

richaas:
Below are the res_ related symbols from libresolv and libc on my debian linux system.
As you can see res_init/ninit/nclose are in libc, whereas res_close lives in libresolv.

An idea would be to use res_ninit()/res_nclose() on all platforms that
have resolv.h, but I need help to get that tested on other platforms.
I've created the branch "resolv" for this purpose. -lresolv is not needed on linux,
but required on mac os x (and possibly other platforms, like solaris).

$ nm -D --defined-only /lib/x86_64-linux-gnu/libresolv.so.2 |grep res_
0000000000007a70 T __libc_res_nquery
00000000000084f0 T __libc_res_nsearch
0000000000007610 T __res_close
00000000000056f0 T __res_dnok
00000000000050b0 T res_gethostbyaddr
0000000000005040 T res_gethostbyname
0000000000004c00 T res_gethostbyname2
00000000000055b0 T __res_hnok
0000000000008350 T __res_hostalias
0000000000007590 T __res_isourserver
0000000000005690 T __res_mailok
0000000000007410 T __res_mkquery
00000000000096d0 T __res_nameinquery
0000000000007770 T __res_nmkquery
00000000000082c0 T __res_nquery
0000000000008300 T __res_nquerydomain
0000000000008b70 T __res_nsearch
000000000000acd0 T __res_nsend
0000000000214200 D _res_opcodes
0000000000005650 T __res_ownok
00000000000097d0 T __res_queriesmatch
00000000000074c0 T __res_query
00000000000076c0 T __res_querydomain
0000000000007630 T __res_search
00000000000075b0 T __res_send
0000000000007550 T res_send_setqhook
0000000000007570 T res_send_setrhook

$ nm -D --defined-only /lib/x86_64-linux-gnu/libc.so.6 |grep res_
00000000003aa800 B _res_hconf
0000000000106500 T __res_iclose
00000000001071d0 T __res_init
0000000000107280 T __res_maybe_init
00000000001065b0 T __res_nclose
00000000001064d0 T __res_ninit
00000000001064e0 T __res_randomid
0000000000107410 T __res_state

czarkoff:
That obviously doesn't work on OpenBSD, as there is no `res_n*` interface here.

richaas:
I've have included a fix for OpenBSD in the resolv branch, please test it. I've also attached a test program to test resolv functionality. Place test.c in your re repo and do a make test to build.
[test.c.txt](https://github.com/creytiv/re/files/523708/test.c.txt)

czarkoff:
That fixes resolver on OpenBSD, but I still don't understand why you would want to use `res_n*` interfaces.

richaas:
Primarily because it is reentrant. As an extra bonus we don't need to link with libresolv on linux. My suggested solution also share the major part of the code between the reentrant and older alternative.

czarkoff:
As far as I can tell your implementation doesn't make use of reentracy.  Furthermore, the very fact that you use `#ifdef` within a single function to choose between old and new API bars you from benefitting of new API in any way.  So the only benefit is linking, and even that seem dubious: from [your comment above](#issuecomment-253018683) it looks like `libresolv` does the actual work anyway, so not linking with it means that you choose a path where it will be dynamically loaded during runtime, which is, in my opinion, something to avoid.

I'd rather drop any support for "new" API.  That would allow keeping source cleaner and reduce the ammount of code paths to tests.  I don't think that ability to avoid explicitly linking against a library that would be loaded anyway is something worth the trouble.

richaas:
We would like to be reentrant on platforms where this is possible. As far as I can see, libresolv is not involved at all (not dynamically loaded) on linux. On the master branch we also have a combination of res_init() and res_ninit(). The latter is currently only used for NetBSD and OpenBSD (where this is not even available!). What I'm proposing is simply to share more code between the two variants and make use of reentrancy on platforms that support res_ninit().

btw: I just successfully tested my resolv branch on OpenBSD

Platforms tested so far: Linux (Debian 8), Mac OS X 10.11 and OpenBSD 5.0.

I would be grateful if we can get this tested on more platforms.

czarkoff:
I tested OpenBSD-current and FreeBSD, both successfully.  Tonight I plan to test NetBSD and DragonFlyBSD.

richaas:
Thanks @czarkoff , much appreciated.

czarkoff:
OK, so NetBSD is fine, while DragonFly obviously isn't.  I isolated necessary changes into another commit, new PR wil follow shortly.

richaas:
Great, I have created a new PR (#11) based on the resolv branch (which now includes DragonFly BSD support). Let's continue the discussion there.

PR #9, opened 2016-10-09
Openssl 1.1.0

alfredh:


alfredh:
the BIO_METHOD is stateless and may be shared by all instances of the TLS-session,
either TCP or UDP. I think we should try to move it away from `struct tls_conn` to e.g.:

1. `struct tls`, the shared TLS context for all sessions
2. static data, but this is not so good for re-entrancy.

alfredh:
please have a look at commit 

https://github.com/creytiv/re/pull/9/commits/13670c824015bc7e060736efce5502d48c741d8e


it moves the creation of BIO_METHOD to the TLS-context:

```
#if OPENSSL_VERSION_NUMBER >= 0x10100000L
	tls->method_tcp = tls_method_tcp();
#endif
```

this approach is slightly more efficient, compared to creating method
for each tls connection.


I have tested this on OSX with openssl 1.1.0 from Homebrew:

```
$ make EXTRA_CFLAGS="-I/usr/local/opt/openssl@1.1/include" EXTRA_LFLAGS="-L/usr/local/opt/openssl@1.1/lib" SYSROOT_ALT= USE_OPENSSL=1 USE_OPENSSL_DTLS=1 USE_DTLS=1 USE_DTLS_SRTP=1 
```

richaas:
Yes, I think this is a better approach, let's try the same for udp.

btw: ERR_clear_error() is not needed by the tls_method_tcp() caller.

alfredh:
the last commit also moves BIO_METHOD for udp to struct tls.

not a super-elegant solution, but it should work okayish for now I think.

czarkoff:
I tried to compile 2b3e0b0 on OpenBSD -current. [It did not go well.](http://pastebin.com/SXCtyEiW)

alfredh:
thanks @czarkoff very good with testing with libressl

I guess we have to find something smarter for the version macro check:

```
/* These will change with each release of LibreSSL-portable */
#define LIBRESSL_VERSION_NUMBER 0x2050000fL
#define LIBRESSL_VERSION_TEXT   "LibreSSL 2.5.0"

/* These will never change */
#define OPENSSL_VERSION_NUMBER  0x20000000L
#define OPENSSL_VERSION_TEXT    LIBRESSL_VERSION_TEXT
#define OPENSSL_VERSION_PTEXT   " part of " OPENSSL_VERSION_TEXT
```

czarkoff:
@alfredh: In my local repo I added ` && !defined(LIBRESSL_VERSION_NUMBER)" to all OpenSSL 1.1.0 tests, and I am still seeing issues.  I think it would probably make more sense to ignore LibreSSL for now and concentrate on OpenSSL.  FWIW I was planning to write a libtls-based implementation of tls for re.

alfredh:
I updated the version check for the new features.

The latest version works fine with openssl 1.0.2, openssl 1.1.0 and libressl 2.4.3

@czarkoff could you please test it again :)

czarkoff:
OpenBSD and FreeBSD build fine, although on FreeBSD compiler complains about implicit declaration:
```
  CC      build-x86_64/tls/openssl/tls.o
src/tls/openssl/tls.c: In function 'tls_srtp_keyinfo':
src/tls/openssl/tls.c:623:2: warning: implicit declaration of function 'SSL_get_selected_srtp_profile' [-Wimplicit-function-declaration]
  sel = SSL_get_selected_srtp_profile(tc->ssl);
  ^
src/tls/openssl/tls.c:623:2: warning: nested extern declaration of 'SSL_get_selected_srtp_profile' [-Wnested-externs]
src/tls/openssl/tls.c:623:6: warning: assignment makes pointer from integer without a cast [enabled by default]
  sel = SSL_get_selected_srtp_profile(tc->ssl);
      ^
```
Indeed grep does not show `SSL_get_selected_srtp_profile` in OpenSSL's headers, although symbol is present in library and linking goes fine.

(This all happens on FreeBSD 11 with OpenSSL 1.0.2j-freebsd.)

alfredh:
rebased against current master.

@richaas Please squash+merge to master.

richaas:
I suggest we try to find a better way to detect libressl; one that don't conflict with future versions of OpenSSL.

czarkoff:
@richaas Pretty reliable way is
```
#ifdef LIBRESSL_VERSION_NUMBER
```

alfredh:
then I can change this:

```
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L && \
+       OPENSSL_VERSION_NUMBER < 0x20000000L
```

to this:

```
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L && \
+       !defined (LIBRESSL_VERSION_NUMBER)
```


while this macro check should work, I think it is starting to become quite
ugly code. Checking for this but not that is not very robust.

do we have any better/clean alternatives?

alfredh:
@richaas I have update the branch to check for libressl version.

richaas:
And then later, when libressl starts to incorporate the openssl-1.1 stuff, this check will break.
I suggest we either drop libressl support altogether or incorporate some sort of libressl to openssl version mapping logic. IMO the first option will do.

alfredh:
I have made a simple feature macro for now, inside the tls module:

```
#define TLS_BIO_OPAQUE 1
```

this makes the code look a bit nicer.

OpenBSD is using libressl by default, and I would like to keep support for
both OpenSSL and LibreSSL.

czarkoff:
FWIW I'd leave out `LIBRESSL_VERSION_NUMBER <= 0x2040300fL` check until LibreSSL actually adds support for that feature: neither LibreSSL 0.2.5 (`0x2050000fL`) nor current revision of LibreSSL support it, and if LibreSSL is going to support it in future, the check could be re-added with appropriate version number.

richaas:
@alfredh I like this approach, and I agree with @czarkoff (I also just checked their dev version 2.5.0, and BIO structs are still open). I suggest we simply update your check like this for now:

#if defined (LIBRESSL_VERSION_NUMBER)
#undef  TLS_BIO_OPAQUE
#endif

@czarkoff, after this is fixed, are there any other libressl related issues left that you are aware of?

czarkoff:
No, I don't think I've seen any other issues with libre and LibreSSL.

alfredh:
@richaas I have made the change as suggested.

I think we should say that this is good enough for now, and merge the branch to master.

please Squash and Merge to master.

src/main/openssl.c
```diff
@@ -18,11 +18,15 @@
...
 
 
-#ifdef HAVE_PTHREAD
+#if defined (HAVE_PTHREAD) && (OPENSSL_VERSION_NUMBER < 0x10100000L)
+
 
 static pthread_mutex_t *lockv;
 
 
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
```
richaas: This is not needed (covered by enclosing #if)

src/main/openssl.c
```diff
@@ -47,6 +51,10 @@ static unsigned long threadid_handler(void)
 #endif
 
 
+#endif  /* OPENSSL_VERSION_NUMBER */
+
+
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
```
richaas: This is not needed (covered by enclosing #if)

czarkoff:
This branch may probably be deleted now...

PR #11, opened 2016-10-15
dns/resolv cleanup

richaas:
General dns/resolv cleanup:

- cleanup resolv support detection
- merge src/dns/res.c and src/dns/bsd/srv.c into one file
- use reentrant versions of res functions where avaliable
- avoid libresolv linking dependency on linux
- add support for DragonFly BSD

This needs testing on more platforms. Tested so far:

- Linux (Debian 8)
- Mac OS X 10.11
- OpenBSD 5.0, 6.0
- FreeBSD
- NetBSD
- DragonFly BSD 

See also #6 for more background.


alfredh:
I have tested this branch on:

* Android NDK (success)
* iOS (success)
* mingw32 (success)

a good reason for using the res_ninit (re-entrant API) is that it supports both
IPv4 and IPv6. The old res_init supports only IPv6.

czarkoff:
> The old res_init supports only IPv4.

Not on BSDs or Linux at least.  From res_init(3) manual page:

> RES_USE_INET6   Enables support for IPv6-only applications.

alfredh:
I think this patch should be good to merge to master :)

richaas:
Would be nice to test on Solaris, do any of you have access to that platform?

czarkoff:
I could try to install it into VM.  In a mean time I found that it breaks Linux with musl libc.

czarkoff:
I tested on Solaris 11.3 on amd64, and it needs `-lresolv`.  When `-lresolv` is provided, it uses reentrant API.   I added a commit to PR #20 (which is what I actually tested).  I plan testing Illumos shortly.

czarkoff:
Tested on Illumos, namely on latest OpenIndiana snapshot.  Again, reentrant API is used, `-lresolv` is needed.

richaas:
@alfredh, this should be good to go now

PR #12, opened 2016-10-16
More diverse testing

czarkoff:
Return OSX tests, and test with clang as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,31 @@
 language: c
 
+os:
+    - linux
+    - osx
+
+compiler:
+    - clang
+    - gcc
+
 before_install:
-    - sudo apt-get -qq update
-    - sudo apt-get -y install libssl-dev
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+        sudo apt-get -qq update;
+        sudo apt-get -y install libssl-dev;
+    fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+        brew update;
+        brew install openssl;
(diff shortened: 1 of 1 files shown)


alfredh:
this patch looks fine to me, ok to merge to master

lgrahl:
To my current knowledge, if-statements in the `.travis.yml` file will prevent catching exit codes of the commands invoked. See [this blog entry](http://steven.casagrande.io/articles/travis-ci-and-if-statements/) for details.

czarkoff:
@lgrahl: I'm afraid you misinterpret the issue:
```
if some test
then
	command1;	/* return value discarded */
	command2;	/* return value discarded */
	command3;	/* return value returned as the value of the whole clause */
fi
```
So in case of single command in if-statement everything is fine.  For reference, if intermediate statements are important, following may be used:
```
if some test
then
	command1 && command2 && command3
fi
```
In this case every any failure will result in non-zero status of if-statement.

lgrahl:
Thanks for your answer. In this case, it's only an issue for the following lines:

```
- if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
      sudo apt-get -qq update;
      sudo apt-get -y install libssl-dev;
  fi
- if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
      brew update;
      brew install openssl;
  fi
```

... which isn't much of an issue since the last commands of the block are vital. :)

czarkoff:
FWIW I simplified preparation stage by switching to `apt` addon.  On OSX all dependencies are already present.

PR #13, opened 2016-10-16
Clean up remnants of Symbian support

czarkoff:
I believe this is all that remained of Symbian in `re`.


alfredh:
I did a quick test on OSX 10.12 with retest, and a quick review.

looks very good to me

PR #14, opened 2016-10-19
Use one DTLS connection for all peers

lgrahl:
These changes make it possible to use one DTLS connection for all peers.
Related: #2

### DIFF
--- a/src/tls/openssl/tls_udp.c
+++ b/src/tls/openssl/tls_udp.c
@@ -26,6 +26,11 @@
 
 
 enum {
+	ONE_PEER_HASH = 0x776f656d
+};
+
+
+enum {
 	MTU_DEFAULT  = 1400,
 	MTU_FALLBACK = 548,
 };
@@ -40,6 +45,7 @@ struct dtls_sock {
 	dtls_conn_h *connh;
 	void *arg;
 	size_t mtu;
+	bool one_peer;
 };
 
 
@@ -343,12 +349,12 @@ static void conn_recv(struct tls_conn *tc, struct mbuf *mb)
 
 			tc->estabh(tc->arg);
 
-                        nrefs = mem_nrefs(tc);
-                        mem_deref(tc);
+			nrefs = mem_nrefs(tc);
+			mem_deref(tc);
 
-                        /* check if connection was deref'd from handler */
-                        if (nrefs == 1)
-                                return;
+			/* check if connection was deref'd from handler */
+			if (nrefs == 1)
+				return;
 		}
 	}
 
@@ -410,13 +416,15 @@ static int conn_alloc(struct tls_conn **ptc, struct tls *tls,
 		      dtls_close_h *closeh, void *arg)
 {
 	struct tls_conn *tc;
+	uint32_t key;
 	int err = 0;
 
 	tc = mem_zalloc(sizeof(*tc), conn_destructor);
 	if (!tc)
 		return ENOMEM;
 
-	hash_append(sock->ht, sa_hash(peer, SA_ALL), &tc->he, tc);
+	key = sock->one_peer ? ONE_PEER_HASH : sa_hash(peer, SA_ALL);
+	hash_append(sock->ht, key, &tc->he, tc);
 
 	tc->sock   = mem_ref(sock);
 	tc->peer   = *peer;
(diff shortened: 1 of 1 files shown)


czarkoff:
You might want to rebase your changes onto [`openssl-1.1.0`](https://github.com/creytiv/re/tree/openssl-1.1.0) branch.

lgrahl:
You mean I should reopen a PR to merge into that branch?

czarkoff:
I merely mean that `openssl-1.1.0` branch is about to be merged, and it changes the same parts of libre as your changes, so you might want to make sure it does not break your changes.

lgrahl:
Ah, okay, I see. Rebase fails with conflicts but recursive merge is looking fine.

src/tls/openssl/tls_udp.c
```diff
@@ -26,6 +26,11 @@
 
 
 enum {
+	ONE_PEER_HASH = 0x776f656d
```
alfredh: the hash value could also be 0x00000000?

alfredh:
a couple of more things to think about..:

- we should perhaps extend the API to update the remote peer of the
  session/connection

src/tls/openssl/tls_udp.c
```diff
@@ -343,12 +349,12 @@ static void conn_recv(struct tls_conn *tc, struct mbuf *mb)
...
-                        mem_deref(tc);
+			nrefs = mem_nrefs(tc);
+			mem_deref(tc);
 
-                        /* check if connection was deref'd from handler */
-                        if (nrefs == 1)
-                                return;
+			/* check if connection was deref'd from handler */
+			if (nrefs == 1)
+				return;
```
lgrahl: Correct.

src/tls/openssl/tls_udp.c
```diff
@@ -721,6 +736,14 @@ int dtls_listen(struct dtls_sock **sockp, const struct sa *laddr,
...
 		goto out;
 
+	if (htsize == 0) {
+		sock->one_peer = true;
+		htsize = 1;
+	}
+	else {
+		sock->one_peer = false;
+	}
+
```
lgrahl: Probably not. Do you want me to look into it?

lgrahl:
> we should perhaps extend the API to update the remote peer of the session/connection

As each `peer` has its own `tls_conn`, would it be sufficient to add a function that changes the `peer` of a specific `tls_conn`?

```c
void dtls_set_peer(struct tls_conn *tc, const struct sa *peer);
```

richaas:
Sending data to TLS connections without any restrictions on the source IP don't
seem like an optimal solution from a security perspective. I think a solution were
we update the peer address on a specfic connection would be better.

lgrahl:
> I think a solution were we update the peer address on a specfic connection would be better.

Maybe I wasn't specific enough: The source address is tied to the `struct dtls_sock`, so the function above would modify the destination address stored in `struct tls_conn`. The source IP wouldn't be altered.

lgrahl:
But isn't `peer` in `struct tls_conn` the other peer's address? Or am I getting something wrong here?

richaas:
I was referring to peer source address.

lgrahl:
I'm not proposing to change that, `dtls_set_peer` would set the other peer's address. Maybe it should be renamed if it's confusing?

richaas:
My comment was about the proposed patch in general, not the dtls_set_peer() function.

lgrahl:
So, basically you'd rather have a function that attaches a peer to an existing connection:

```c
void dtls_add_peer(struct tls_conn *tc, const struct sa *peer);
```

Which is a bit tricky because AFAIK an existing `struct tls_conn` cannot be re-added to the hash list with another peer address.

What we could do is introducing an intermediate struct:

```c
struct tls_conn_peer {
    struct le he;
    struct sa peer;
    struct tls_conn* conn;
}
```

The first peer added to a `struct tls_conn` would set the `peer` member to its address. `conn_lookup` would look up the peer from the hash list of `struct tls_conn_peer`s. When sending, the `struct tls_conn.peer` member would still be used as destination address.

Opinions?

lgrahl:
If there's nothing standing against the proposal above, I'll write the code for it and open up an alternative PR.

richaas:
AFAIU, we only need one peer at any given time. If so, then I suggest we simply update the peer IP address when needed.

lgrahl:
This is not the case for [ICE](https://tools.ietf.org/html/draft-ietf-ice-rfc5245bis-04#section-9.2) where we...

> MUST be prepared to receive media on each
component on any candidates provided for that component in the most
recent candidate exchange

So, if I interpret this correctly, I must be prepared to receive DTLS messages on any of the remote candidates. And even if it wouldn't be a *must*, I would still like to add the remote address as some implementations may choose to send it from an arbitrary local candidate's address.

richaas:
Yes, but you still have only one "active" peer at any given point in time, even if the address/port may change during the session/ICE setup. Using our UDP helper interface, its possible to intercept
UDP packets before they reach the DTLS layer. Here you can for example update the DTLS peer as needed, based on some application specific policy.

lgrahl:
True indeed. I'll see if I can get it to work with updating the peer address only and come back with another PR to update the peer's address.

lgrahl:
Thanks @richaas, your suggestion works fine for my use case and it's much simpler. I'll close this one in favour of #24

PR #15, opened 2016-10-20
Setting Certificate/Private Key Improvements

lgrahl:
These changes give the user a little bit more fine granularity when setting a certificate (and the private key).
Other private key types than RSA are now supported.
PEM certificate and private key can now be contained in separate strings.
Moreover, the user is able to use the DER format for certificates and private keys.

- `tls_set_certificate_pem`
- `tls_set_certificate_der`
- for compatibility, `tls_set_certificate` has been updated and uses `tls_set_certificate_pem` underneath

No changes to existing function signatures had to be made. :)

(Let me know if I should rebase the changes and make a PR to merge into the `openssl-1.1.0` branch.)

### DIFF
--- a/include/re_tls.h
+++ b/include/re_tls.h
@@ -24,11 +24,21 @@ enum tls_fingerprint {
 	TLS_FINGERPRINT_SHA256,
 };
 
+enum tls_key_type {
+	TLS_KEY_TYPE_RSA,
+	TLS_KEY_TYPE_EC,
+};
+
 
 int tls_alloc(struct tls **tlsp, enum tls_method method, const char *keyfile,
 	      const char *pwd);
 int tls_add_ca(struct tls *tls, const char *capath);
 int tls_set_selfsigned(struct tls *tls, const char *cn);
+int tls_set_certificate_pem(struct tls *tls, const char *cert, size_t len_cert,
+		const char *key, size_t len_key);
+int tls_set_certificate_der(struct tls *tls, enum tls_key_type key_type,
+		const uint8_t *cert, size_t len_cert, const uint8_t *key,
+		size_t len_key);
 int tls_set_certificate(struct tls *tls, const char *cert, size_t len);
 void tls_set_verify_client(struct tls *tls);
 int tls_set_srtp(struct tls *tls, const char *suites);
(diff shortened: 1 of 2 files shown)


lgrahl:
Note: I've double-checked that the newly introduced OpenSSL functions are available in 0.9.8 to 1.1.0 and libressl. I've compiled and tested it on OpenSSL 1.0.1f and OpenSSL 1.0.2h.

alfredh:
thanks for this nice patch. in general all the code looks of superb quality and
should be good to merge to master.

in general all patches/PRs should be based on master. the order in which we merge
patches/PRs to master can sometimes be random.

it was particularly good that you added this type:

```
enum tls_key_type {
       TLS_KEY_TYPE_RSA,
       TLS_KEY_TYPE_EC,
};
```

later we should extend the selfsign function, to include keysize and also
add support for Elliptic Curve (EC).

lgrahl:
Thanks for the feedback. Much appreciated. :)

> later we should extend the selfsign function, to include keysize and also
add support for Elliptic Curve (EC).

Already wrote some code for that in my own library. But if you're interested in it, I can easily port it and open a PR.

richaas:
Hi @lgrahl, would it be possible to have a new patch here, were you keep the original variable names from tls_set_certificate (except for the rsa/pkey variable). The diff would be much cleaner/easier to read then.

Also, would it be possible to move the DER related changes to another PR? BTW, in your DER code it looks like you reference updated buf_cert value from d2i_X509(), before checking d2i_X509() return value.

lgrahl:
> Hi @lgrahl, would it be possible to have a new patch here, were you keep the original variable names from tls_set_certificate (except for the rsa/pkey variable). The diff would be much cleaner/easier to read then.

Don't get me wrong, but I don't think changing the variables names back to make the diff nicer to read but overall have confusing variable names as a result is a good idea.

> Also, would it be possible to move the DER related changes to another PR?

I could do that but I would prefer having it this way unless there is a reason why I should split them up. Also, the function name `tls_set_certificate_pem` doesn't make much sense without the `der` counterpart.

> BTW, in your DER code it looks like you reference updated buf_cert value from d2i_X509(), before checking d2i_X509() return value.

Nice find, will fix that.

lgrahl:
Closing in favour of #25, @richaas #25 also fixes the DER function and addresses your request regarding variable names.

PR #20, opened 2016-10-25
Detect resolver API based on __RES version

czarkoff:
`__RES` is resolver version expressed as date.  It allows rather reliably determine whether implementation supports reentrant API or not.  If implementation does not define `__RES`, assume reentrant API.


richaas:
Thanks @czarkoff, the -lresolv for Solaris part is in resolv branch now.

czarkoff:
@richaas, could you please comment on `__RES` commit?

PR #21, opened 2016-10-27
Unbreak re on Oracle Solaris Studio

czarkoff:
This PR contains changes I had to make to my local tree while trying to compile re on Solaris 11.3 with Oracle Solaris Studio 12.5.

PR #22, opened 2016-10-27
Fixes for Solaris 11 and Illumos (Solaris 10)

czarkoff:
These commits clean up some warnings when compiling libre on Illumos distributions and on Oracle Solaris 11 using gcc.  The first three of them also allow using Solaris Studio cc, which currently can't successfully finish the build.  Otherwise they should not affect other platforms, and they seem to be harmless on Linux (as in Travis), FreeBSD and OpenBSD.  Tested on latest snapshot of OpenIndiana (with gcc) and Oracle Solaris 11.3 (with gcc and Solaris Studio cc).

P.S.:  This is basically a cleaned up version of PR #21.

### DIFF
--- a/src/fmt/pl.c
+++ b/src/fmt/pl.c
@@ -6,7 +6,7 @@
 #include <ctype.h>
 #include <sys/types.h>
 #ifdef HAVE_STRINGS_H
-#define __EXTENSIONS__ 1
+#define _XPG4_2
 #include <strings.h>
 #endif
 #include <string.h>
--- a/src/main/main.c
+++ b/src/main/main.c
@@ -155,11 +155,7 @@ static struct re global_re = {
 
 static void poll_close(struct re *re);
 
-#ifdef SOLARIS
-static pthread_once_t pt_once = {PTHREAD_ONCE_INIT};
-#else
 static pthread_once_t pt_once = PTHREAD_ONCE_INIT;
-#endif
 static pthread_key_t  pt_key;
 
 
--- a/mk/re.mk
+++ b/mk/re.mk
@@ -182,10 +182,12 @@ endif
 endif
 
 # Compiler dependency flags
+ifeq ($(CC_NAME),$(filter $(CC_NAME),gcc clang))
 ifeq ($(CC_SHORTVER), 2.9x)
 	DFLAGS		= -MD
 else
 	DFLAGS		= -MD -MF $(@:.o=.d) -MT $@
+endif
 endif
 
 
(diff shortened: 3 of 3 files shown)


src/fmt/pl.c
```diff
@@ -6,7 +6,7 @@
 #include <ctype.h>
 #include <sys/types.h>
 #ifdef HAVE_STRINGS_H
-#define __EXTENSIONS__ 1
+#define _XPG4_2
```
alfredh: suggest that we keep `__EXTENSIONS__`

src/main/main.c
```diff
@@ -155,11 +155,7 @@ static struct re global_re = {
 
 static void poll_close(struct re *re);
 
-#ifdef SOLARIS
-static pthread_once_t pt_once = {PTHREAD_ONCE_INIT};
-#else
 static pthread_once_t pt_once = PTHREAD_ONCE_INIT;
-#endif
 static pthread_key_t  pt_key;
```
alfredh: I think I added this ifdef SOLARIS thing a long time ago. In general this hunk is a good one.

src/fmt/pl.c
czarkoff: If `__EXTENSIONS__` defined, `strncasecmp()` is hidden on Illumos.  See their [strings.h](https://github.com/illumos/illumos-gate/blob/f7877f5d39900cfd8b20dd673e5ccc1ef7cc7447/usr/src/head/strings.h).

src/main/main.c
czarkoff: It makes the Solaris 11 build fail.  I did not try to build it on Illumos, but it seems fine without it.  It might have been needed on some really ancient version of Solaris, who knows...

alfredh:
in general we should support the most common version of Solaris with the default
toolchain/compiler.

Illumos is not supported.

in general I am not interested in supporting a bunch of rare and unpopular platforms.
supporting many platforms has a certain cost, for every release we need to compile and
test that things are working.

richaas:
I agree with @alfredh in his above comment

czarkoff:
Could you please clarify: what is, in your opinion, "the most common version of Solaris with the default toolchain/compiler"?

alfredh:
Hey @czarkoff 

if you have access to a Solaris box, could you please try to compile libre@HEAD
with gcc and paste the complete build log here?

$ make CC=gcc


this used to work in the past (i.e. a couple of years ago :)


I guess the current version of Solaris as distributed by Oracle is the version
that we should support ? Perhaps some people are running older versions
of Solaris, but I guess these versions are not supported by Oracle either?

czarkoff:
I will deploy another test vm shortly and provide a log.  Which version of GCC do you want me to use - GCC is not the default toolchain, so I need to pick a version to install.

Regarding "main" Solaris version.  It seems that most people either use whatever Solaris version that came with their branded hardware or Illumos if they build their own servers independently.  Apparently that should mean that Solaris 10 (either SUN-branded or Illumos distributions) should be the average system these days.

At any rate, if Sun Studio is not supported, and there will be no accommodations for it, I guess at very least the corresponding fragment of `mk/re.mk` should be removed, so that people like me don't waste inadequate amount of time on fixing it.

czarkoff:
[Log for re on Solaris 11.3 with GCC 4.8.](http://pastebin.com/hGVtD2Xv)

alfredh:
I have made an alternative patch that fixes Solaris 11 with gcc:

https://github.com/creytiv/re/pull/52

alfredh:
hi @czarkoff the latest version of libre works fine with Solaris 11.3 and gcc.

I have tested with Baresip and Retest.

PR #23, opened 2016-10-28
Detect resolver API based on __RES version

czarkoff:
As BIND's `include/resolv.h` puts it:

> Revision information.  This is the release date in YYYYMMDD format.
> It can change every day so the right thing to do with it is use it
> in preprocessor commands such as "#if (__RES > 19931104)".  Do not
> compare for equality; rather, use it to determine whether your resolver
> is new enough to contain a certain feature.

Reentrant resolver API was introduced in BIND 8.2, which set `__RES` to `19980901`.
Thus, testing for

```
#if __RES >= 19980901
```

provides a reliable way to test for availability of reentrant resolver API.

In my testing this gives appropriate results across BSDs (FreeBSD, DragonFly BSD, NetBSD and OpenBSD), Linux with GNU and musl libc, Illumos and Solaris 11.  According to Travis, it builds without issues on macOS as well.  Of these platforms all but OpenBSD and Linux+musl provide reentrant resolver API _and_ define `__RES` to value higher then `19980901`.

If this change is not wanted for some reason, I would appreciate explanation.

### DIFF
--- a/src/dns/res.c
+++ b/src/dns/res.c
@@ -25,12 +25,12 @@ int get_resolv_dns(char *domain, size_t dsize, struct sa *nsv, uint32_t *n)
 	uint32_t i;
 	int ret, err;
 
-#ifdef OPENBSD
-	ret = res_init();
-	state = _res;
-#else
+#if __RES >= 19980901
 	memset(&state, 0, sizeof(state));
 	ret = res_ninit(&state);
+#else
+	ret = res_init();
+	state = _res;
 #endif
 	if (0 != ret)
 		return ENOENT;
@@ -56,8 +56,7 @@ int get_resolv_dns(char *domain, size_t dsize, struct sa *nsv, uint32_t *n)
 	*n = i;
 
  out:
-#ifdef OPENBSD
-#else
+#if __RES >= 19980901
 	res_nclose(&state);
 #endif
 
--- a/src/dns/res.c
+++ b/src/dns/res.c
@@ -59,6 +59,8 @@ int get_resolv_dns(char *domain, size_t dsize, struct sa *nsv, uint32_t *n)
  out:
 #if __RES >= 19980901
 	res_nclose(&state);
+#elif !defined(OPENBSD)
+	res_close();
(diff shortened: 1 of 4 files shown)


alfredh:
in general this change look ok. I will try to test it on a couple of
different platforms..

src/dns/res.c
```diff
@@ -25,12 +25,12 @@ int get_resolv_dns(char *domain, size_t dsize, struct sa *nsv, uint32_t *n)
 	uint32_t i;
 	int ret, err;
 
-#ifdef OPENBSD
-	ret = res_init();
-	state = _res;
-#else
+#if __RES >= 19980901
```
alfredh: could you please add a small comment above this ifdef, such as:

```Reentrant resolver API was introduced in BIND 8.2, which set __RES to 19980901.```

richaas:
Please note that OpenBSD lacks the res_close() function, whereas other, non-reentrant systems likely have this function. I would therefor like to keep #ifdef OPENBSD, at least for the close section.

czarkoff:
FWIW i would consider going in opposite direction: in non-reentrant case initialize only once and never close.

czarkoff:
ping

alfredh:
Hey @czarkoff 

the current approach works quite well and has by now been tested on many platforms
and deployed into running systems.

as your patch does not address a concrete problem, I would like to suggest that
we close this for now. If this really becomes an issue later, we should revisit it.

/Alfred

czarkoff:
@alfredh This fixes on of two immediate issues with `re` on musl libc and makes porting `re` to other unsupported platforms a bit easier.  Provided that for two months of this PR's existence nobody managed to come up with any argument against it, and that this detection mechanism is - as opposed to the current one - correct, make me think that it would make much more sense to merge it.  But it is not for me to decide, obviously.

alfredh:
@czarkoff your contributions are very useful and we appreciate it alot.

But please understand that we do not want to merge this particular PR.

please close the PR for now.


/Alfred

PR #24, opened 2016-10-31
Change DTLS connection's remote peer address

lgrahl:
These changes add the possibility to get and set the DTLS connection's remote peer address.

```c
const struct sa *dtls_peer(const struct tls_conn *tc);
void dtls_set_peer(struct tls_conn *tc, const struct sa *peer);
```

### DIFF
--- a/src/tls/openssl/tls_udp.c
+++ b/src/tls/openssl/tls_udp.c
@@ -652,7 +652,7 @@ void dtls_set_handlers(struct tls_conn *tc, dtls_estab_h *estabh,
 
 
 /**
- * Get the remote peer a DTLS Connection
+ * Get the remote peer of a DTLS Connection
  *
  * @param tc DTLS Connection
  *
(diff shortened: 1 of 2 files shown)


richaas:
@lgrahl, glad to hear this approach works for you. API wise, I was thinking something like:

void dtls_set_peer(struct tls_conn *tc, const struct sa *peer);
const struct sa *dtls_peer(const struct tls_conn *tc);

to set and get peer address respectively. Without and sa comparison done inside dtls_set_peer().
How would that fit your use case?

lgrahl:
That would be fine as well. I'll update the PR tomorrow.

lgrahl:
@richaas Updated the PR with your suggestions. Also updated the description of the first posting.

src/tls/openssl/tls_udp.c
```diff
@@ -651,6 +651,37 @@ void dtls_set_handlers(struct tls_conn *tc, dtls_estab_h *estabh,
 }
 
 
+/**
+ * Get the remote peer a DTLS Connection
```
richaas: Get the remote peer _of_ a DTLS connection

lgrahl: Fixed in 101727c6f12bd1718c967bb9c3a05a8654a61237

richaas:
thanks @lgrahl

@alfredh feel free to merge if you agree with these changes.

PR #25, opened 2016-11-01
Setting Certificate/Private Key Improvements

lgrahl:
These changes give the user a little bit more fine granularity when setting a certificate (and the private key).
Other private key types than RSA are now supported.
PEM certificate and private key can now be contained in separate strings.
Moreover, the user is able to use the DER format for certificates and private keys.

- `tls_set_certificate_pem`
- `tls_set_certificate_der`
- for compatibility, `tls_set_certificate` has been updated and uses `tls_set_certificate_pem` underneath

No changes to existing function signatures had to be made. :)

(Rebased and fixed version as a replacement for #15)

### DIFF
--- a/include/re_tls.h
+++ b/include/re_tls.h
@@ -24,9 +24,9 @@ enum tls_fingerprint {
 	TLS_FINGERPRINT_SHA256,
 };
 
-enum tls_key_type {
-	TLS_KEY_TYPE_RSA,
-	TLS_KEY_TYPE_EC,
+enum tls_keytype {
+	TLS_KEYTYPE_RSA,
+	TLS_KEYTYPE_EC,
 };
 
 
@@ -36,7 +36,7 @@ int tls_add_ca(struct tls *tls, const char *capath);
 int tls_set_selfsigned(struct tls *tls, const char *cn);
 int tls_set_certificate_pem(struct tls *tls, const char *cert, size_t len_cert,
 		const char *key, size_t len_key);
-int tls_set_certificate_der(struct tls *tls, enum tls_key_type key_type,
+int tls_set_certificate_der(struct tls *tls, enum tls_keytype key_type,
 		const uint8_t *cert, size_t len_cert, const uint8_t *key,
 		size_t len_key);
 int tls_set_certificate(struct tls *tls, const char *cert, size_t len);
--- a/include/re_tls.h
+++ b/include/re_tls.h
@@ -24,11 +24,21 @@ enum tls_fingerprint {
 	TLS_FINGERPRINT_SHA256,
 };
 
+enum tls_key_type {
+	TLS_KEY_TYPE_RSA,
+	TLS_KEY_TYPE_EC,
+};
+
 
 int tls_alloc(struct tls **tlsp, enum tls_method method, const char *keyfile,
 	      const char *pwd);
 int tls_add_ca(struct tls *tls, const char *capath);
 int tls_set_selfsigned(struct tls *tls, const char *cn);
+int tls_set_certificate_pem(struct tls *tls, const char *cert, size_t len_cert,
+		const char *key, size_t len_key);
+int tls_set_certificate_der(struct tls *tls, enum tls_key_type key_type,
+		const uint8_t *cert, size_t len_cert, const uint8_t *key,
+		size_t len_key);
 int tls_set_certificate(struct tls *tls, const char *cert, size_t len);
 void tls_set_verify_client(struct tls *tls);
 int tls_set_srtp(struct tls *tls, const char *suites);
--- a/src/tls/openssl/tls.c
+++ b/src/tls/openssl/tls.c
@@ -74,15 +74,15 @@ static int password_cb(char *buf, int size, int rwflag, void *userdata)
 }
 
 
-static int key_type2int(enum tls_key_type type)
+static int keytype2int(enum tls_keytype type)
 {
 	switch (type) {
-		case TLS_KEY_TYPE_EC:
-			return EVP_PKEY_EC;
-		case TLS_KEY_TYPE_RSA:
-			return EVP_PKEY_RSA;
-		default:
-			return EVP_PKEY_NONE;
+	case TLS_KEYTYPE_EC:
+		return EVP_PKEY_EC;
+	case TLS_KEYTYPE_RSA:
+		return EVP_PKEY_RSA;
+	default:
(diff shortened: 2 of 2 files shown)


src/tls/openssl/tls.c
```diff
@@ -396,6 +406,101 @@ int tls_set_certificate(struct tls *tls, const char *pem, size_t len)
 	return err;
 }
 
+/**
+ * Set the certificate and private key on a TLS context
+ *
+ * @param tls      TLS Context
+ * @param cert     Certificate in DER format
```
richaas: Use same order of parameter as signature

src/tls/openssl/tls.c
```diff
@@ -396,6 +406,101 @@ int tls_set_certificate(struct tls *tls, const char *pem, size_t len)
...
+		EVP_PKEY_free(pkey);
+	if (err)
+		ERR_clear_error();
+
+	return err;
+}
+
+/**
+ * Set the certificate and private key on a TLS context
+ * @deprecated Use tls_set_certificate_pem or tls_set_certificate_der instead
```
richaas: No need to mark this as _deprecated_

src/tls/openssl/tls.c
```diff
@@ -337,55 +339,63 @@ int tls_set_selfsigned(struct tls *tls, const char *cn)
...
+	x509 = NULL;
 
 	err = 0;
 
- out:
-	if (cert)
-		X509_free(cert);
-	if (rsa)
-		RSA_free(rsa);
+out:
```
richaas: keep space before label

src/tls/openssl/tls.c
```diff
@@ -396,6 +406,101 @@ int tls_set_certificate(struct tls *tls, const char *pem, size_t len)
...
+
+	x509 = d2i_X509(NULL, &buf_cert, len_cert);
+	if (!x509)
+		goto out;
+	if (!key) {
+		buf_key = buf_cert;
+		len_key = len_cert - (buf_cert - cert);
+	}
+	else {
+		buf_key = key;
```
richaas: variable buf_key is not needed, use key directly

src/tls/openssl/tls.c
```diff
@@ -396,6 +406,101 @@ int tls_set_certificate(struct tls *tls, const char *pem, size_t len)
...
+{
+	const uint8_t *buf_cert, *buf_key;
+	X509 *x509 = NULL;
+	EVP_PKEY *pkey = NULL;
+	int r, type, err = ENOMEM;
+
+	if (!tls || !cert || !len_cert || (key && !len_key))
+		return EINVAL;
+
+	switch (key_type) {
```
richaas: move this into a separate mapping function, as this can be useful for several functions in the future:

`static int keytype(enum tls_keytype type)`

src/tls/openssl/tls.c
```diff
@@ -396,6 +406,101 @@ int tls_set_certificate(struct tls *tls, const char *pem, size_t len)
...
+
+	if (tls->cert)
+		X509_free(tls->cert);
+
+	tls->cert = x509;
+	x509 = NULL;
+
+	err = 0;
+
+out:
```
richaas: space before label

src/tls/openssl/tls.c
```diff
@@ -337,55 +339,63 @@ int tls_set_selfsigned(struct tls *tls, const char *cn)
...
 		return EINVAL;
 
-	bio  = BIO_new_mem_buf((char *)pem, (int)len);
-	kbio = BIO_new_mem_buf((char *)pem, (int)len);
+	if (!key) {
+		key = cert;
+		len_key = len_cert;
+	}
+
+	bio = BIO_new_mem_buf((char *)cert, (int)len_cert);
```
richaas: keep double space after bio, to simplify diff

src/tls/openssl/tls.c
lgrahl: Oh... I didn't even notice that. Out of curiosity: Any particular reason why you're adding a space before a label?

src/tls/openssl/tls.c
richaas: I'm aware that the pointer is updated, but this is not an issue in our use case.

src/tls/openssl/tls.c
richaas: just our coding standard

src/tls/openssl/tls.c
lgrahl: Because? The caller doesn't expect `key` to be advanced.

richaas: caller will not be affected; caller pass *, not **.

include/re_tls.h
```diff
@@ -24,11 +24,21 @@ enum tls_fingerprint {
 	TLS_FINGERPRINT_SHA256,
 };
 
+enum tls_key_type {
```
richaas: Please use naming enum tls_keytype (TLS_KEYTYPE_RSA, TLS_KEYTYPE_EC)

src/tls/openssl/tls.c
lgrahl: D'oh... my mistake.

src/tls/openssl/tls.c
```diff
@@ -72,6 +74,19 @@ static int password_cb(char *buf, int size, int rwflag, void *userdata)
 }
 
 
+static int key_type2int(enum tls_key_type type)
+{
+	switch (type) {
+		case TLS_KEY_TYPE_EC:
```
richaas: fix indentation inside switch

alfredh:
all these changes looks good.

I have tested that it works with `retest`: https://github.com/creytiv/retest

```
$ ./retest -r
retest: libre version 0.5.0 (x86_64/darwin)
using async polling method 'kqueue'
regular tests:       json: value of unknown type: <t>
json: value of unknown type: <f>
json: value of unknown type: <n>
json: value of unknown type: <frue>
json: value of unknown type: <talse>
uric: unescape: short uri (0)
uric: unescape: short uri (0)
OK
```

perhaps we should update the TLS test to use a DER-encoded
cert, or a PEM with a ECDSA key?

alfredh:
extra testing after merge (718583f8a96ce0dfcea8a0f51ff19a4e3c7565e3)

* Mingw32 with OpenSSL 1.1.0b
* Android NDK r13 with OpenSSL 1.0.2j

PR #26, opened 2016-11-06
Readme

alfredh:
this patch moves the README file from docs directory to the root, and changes
the format to markdown so that it can be displayed on Github homepage.


czarkoff:
@alfredh, new README.md could use some more formatting.  See PR #27 for example.

PR #35, opened 2016-11-22
Set Diffie-Hellman parameters

lgrahl:
Add the possibility to set Diffie-Hellman parameters for DH(E) and ECDH(E).

The following functions have been added:

* `tls_set_dh_params_pem` to read DH parameters from a PEM string
* `tls_set_dh_params_der` to read DH parameters from DER-encoded binary

When setting Diffie-Hellman parameters, those parameters will also be added for elliptic curve ciphers.
For older OpenSSL versions that do not support the `SSL_CTX_set_ecdh_auto` function, only the NIST P-256 curve is being enabled for ECDH(E).

### DIFF
--- a/src/tls/openssl/tls.c
+++ b/src/tls/openssl/tls.c
@@ -10,6 +10,8 @@
 #include <openssl/bn.h>
 #include <openssl/evp.h>
 #include <openssl/x509.h>
+#include <openssl/dh.h>
+#include <openssl/ec.h>
 #include <re_types.h>
 #include <re_fmt.h>
 #include <re_mem.h>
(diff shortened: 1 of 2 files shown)


src/tls/openssl/tls.c
```diff
@@ -847,6 +849,157 @@ int tls_set_ciphers(struct tls *tls, const char *cipherv[], size_t count)
...
+	ec_key = EC_KEY_new_by_curve_name(NID_X9_62_prime256v1);
+	if (!ec_key)
+		return ENOMEM;
+	r = SSL_CTX_set_tmp_ecdh(tls->ctx, ec_key);
+	EC_KEY_free(ec_key);
+	if (!r) {
+		DEBUG_WARNING("set_dh_params: set_tmp_ecdh failed\n");
+		return ENOMEM;
+	}
+#endif
```
alfredh: I think we should move this block to a separate function, something like `tls_enable_ecdh`.

in the case of non-auto, would it make sense to take in curve name as an input parameter
to the function ?

alfredh:
Hi @lgrahl many thanks for your patch.

I have a couple of comments:

* would it be possible for you to add a testcase for this in `retest`? when it comes to crypto
it is quite useful to have a testcase than can be tested on many different platforms.
the place to add it would be in either tls.c or dtls.c, which creates a TLS loop (client+server)
and verifies that a TLS handshake is working. I think the PEM case would be enough.

* I am wondering if we should extract the tmp ecdh block to its own function. This can be useful
to call in other cases too.

I have been using tlsperf (https://github.com/alfredh/tlsperf) to test TLS performance
and other things. Using openssl 1.0.2 and 1.1.0 I see that the negotiated cipher
is different:

openssl 1.0.2:
```
$ ./tlsperf -n 1
tlsperf -- TLS performance testing program
build:         64-bit little endian
compiler:      4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)
libre:         0.5.0
os:            darwin
arch:          x86_64
openssl info:  OpenSSL 1.0.2j  26 Sep 2016
compiler: /usr/bin/clang -I. -I.. -I../include  -fPIC -fno-common -DOPENSSL_PIC -DZLIB -DOPENSSL_THREADS -D_REENTRANT -DDSO_DLFCN -DHAVE_DLFCN_H -arch x86_64 -O3 -DL_ENDIAN -Wall -DOPENSSL_IA32_SSE2 -DOPENSSL_BN_ASM_MONT -DOPENSSL_BN_ASM_MONT5 -DOPENSSL_BN_ASM_GF2m -DSHA1_ASM -DSHA256_ASM -DSHA512_ASM -DMD5_ASM -DAES_ASM -DVPAES_ASM -DBSAES_ASM -DWHIRLPOOL_ASM -DGHASH_ASM -DECP_NISTZ256_ASM
protocol:      TLS
certificate:   selfsigned RSA-1024
starting tests now. (num=1)
[ 1 .. ! ]
DONE!
cipher:        AES256-GCM-SHA384
~~~ Summary: ~~~
num_connections:      1
total_duration:       2 ms
avg_time_per_conn:    2.000 ms
connections_per_sec:  500.000

```

openssl 1.1.0c:
```
$ ./tlsperf -n 1
tlsperf -- TLS performance testing program
build:         64-bit little endian
compiler:      6.2.1 20161124
libre:         0.5.0
os:            linux
arch:          x86_64
openssl info:  OpenSSL 1.1.0c  10 Nov 2016
compiler: gcc -DDSO_DLFCN -DHAVE_DLFCN_H -DNDEBUG -DOPENSSL_THREADS -DOPENSSL_NO_STATIC_ENGINE -DOPENSSL_PIC -DOPENSSL_IA32_SSE2 -DOPENSSL_BN_ASM_MONT -DOPENSSL_BN_ASM_MONT5 -DOPENSSL_BN_ASM_GF2m -DSHA1_ASM -DSHA256_ASM -DSHA512_ASM -DRC4_ASM -DMD5_ASM -DAES_ASM -DVPAES_ASM -DBSAES_ASM -DGHASH_ASM -DECP_NISTZ256_ASM -DPOLY1305_ASM -DOPENSSLDIR="\"/usr/lib/ssl\"" -DENGINESDIR="\"/usr/lib/x86_64-linux-gnu/engines-1.1\"" 
protocol:      TLS
certificate:   selfsigned RSA-1024
starting tests now. (num=1)
[ 1 .. ! ]
DONE!
cipher:        ECDHE-RSA-AES256-GCM-SHA384
~~~ Summary: ~~~
num_connections:      1
total_duration:       2 ms
avg_time_per_conn:    2.000 ms
connections_per_sec:  500.000

```

this is with the same version of libre and tlsperf. Openssl 1.0.2 chooses
AES256-GCM-SHA384 (no PFS) and OpenSSL 1.1.0 chooses
ECDHE-RSA-AES256-GCM-SHA384 (PFS enabled).


I am happy to help with more testing of this patch :)

src/tls/openssl/tls.c
lgrahl: I'm not sure if enabling ECDH depends on having set DH parameters. I'll need to look into that. Furthermore, newer OpenSSL versions have ECDH enabled automatically and `SSL_CTX_set_ecdh_auto` is a no-op (again, not sure if that depends on having set DH params). We could move this to a separate function but that would mean that ECDH might still be enabled although the `tls_enable_ecdh` function hasn't been called which could be confusing to the user. What do you think?

lgrahl:
1. Sure, I'll add a PR to `retest` (fyi, I've also started writing tests for #25 but it isn't ready yet). I'm a bit busy at the moment so this could take a bit of time. If anyone else wants to jump in, feel free to do so. :)
2. I've added a comment above regarding to move code into `tls_enable_ecdh`.
3. Have you added the changes of this PR to `tlsperf` and you got the results above or was this without the changes of this PR?

alfredh:
from https://www.openssl.org/news/cl110.txt:

```
  *) SSL_{CTX_}set_ecdh_auto() has been removed and ECDH is support is
     always enabled now.  If you want to disable the support you should
     exclude it using the list of supported ciphers. This also means that the
     "-no_ecdhe" option has been removed from s_server.
     [Kurt Roeckx]
```

I think now that ECDH should be enabled by default in tls_alloc. This means
that the libre API is consistent, independently of which OpenSSL version is used.

I have been testing this in the past, and it was working except on RedHat based
platforms like CentOS and Fedora, because they ship with a version of OpenSSL
that contains very few curves. On platforms where it was working, I did not have
to set DH parameters explicitly.

alfredh:
@lgrahl is this PR still valid ?

lgrahl:
Yes. Sorry for the delay, I will come back to this after a brief holiday. :)

alfredh:
thank you. I will also try to prepare the ECDH stuff as a separate patch.

richaas:
Instead of adding this to libre, I suggest we add an accessor function returning the OpenSSL context, allowing the application to do things like this without having to modify libre.

alfredh:
I think this is a good idea to add accessors to openssl.

in the past I have also had similar usecases, that would have been solved with accessor.

if you want I can try to create a PR for this.

richaas:
Sounds good, please go ahead.

alfredh:
@lgrahl I will close this issue for now.

You can now use this accessor in your code to modify the OpenSSL context:

https://github.com/creytiv/re/blob/master/include/re_tls.h#L95

PR #36, opened 2016-12-23
HTTP client improvements

richaas:
This PR adds the following improvements to the HTTP client:
* TCP connection reuse
* Ability to deref (abort) request from the application data handler
* Optional request reqp argument (pending requests are kept in client linked list)

@alfredh please test, review and apply.

### DIFF
--- a/src/http/client.c
+++ b/src/http/client.c
@@ -10,6 +10,7 @@
 #include <re_mbuf.h>
 #include <re_sa.h>
 #include <re_list.h>
+#include <re_hash.h>
 #include <re_fmt.h>
 #include <re_tmr.h>
 #include <re_srtp.h>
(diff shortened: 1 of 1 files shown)


src/http/client.c
```diff
@@ -22,43 +23,74 @@
...
 
 
+struct conn {
+	struct tmr tmr;
+	struct sa addr;
+	struct le he;
+	struct http_req *req;
+	struct tls_conn *sc;
+	struct tcp_conn *tc;
+	uint64_t usec;
```
alfredh: when I first read this line I was thinking "micro-seconds"

when reading the code I found out it meant "usecount".

perhaps rename to "use_count" and use a shorter datatype?

src/http/client.c
```diff
@@ -198,24 +317,94 @@ static void recv_handler(struct mbuf *mb, void *arg)
...
+	const struct conn *conn = le->data;
+	const struct http_req *req = arg;
+
+	if (!sa_cmp(&req->srvv[req->srvc], &conn->addr, SA_ALL))
+		return false;
+
+	if (req->secure != !!conn->sc)
+		return false;
+
+	return conn->req == NULL;
```
alfredh: I guess conn->req is set if it has a pending request, and conn->req==NUL if the connection
is free to be reused by other requests?

perhaps we should add a small comment about this.

alfredh:
some comments in the code, more comments here:

* I am wondering if the client should read the `Connection:` header field and determine
if the connection can be reused. For example `Connection: close` from the server signals
that the connection can NOT be reused. Also `Connection: upgrade` as used by
Websockets I guess cannot be reused.

* when writing a test-client that implements both http_resp_handler and http_data_handler,
I can see that the resp_handler is called twice:

1. resp(0, msg, arg)
2. data(mbuf, arg)
3. resp(0, 0, arg)

is this correct, and is the same behaviour as in master?

richaas:
Yes, this is correct. When the data handler is set, the resp handler is called after end of transfer (err=0), or if an error occurs (err!=0). The client stack need to keep track of transfer progress, in order to know that the connection is free for reuse.

richaas:
In the websocket case, the request will be deref'ed in the response handler, and the connection with it. The connection is pure websocket from that point (Connection: upgrade).

richaas:
Added code to handle `Connection: close` responses.

alfredh:
I have tested this version with a REST client against a REST server with
quite heavy traffic. It works as expected, and several TCP-connections are reused.

while the client is sending hundreds of HTTP requests to the server, the number
of simultaneous TCP-connections varies between 1-6. After one minute or so the
TCP-connection is closed by the server, forcing the client to open a new connection.

PR #41, opened 2017-02-15
srtcp: use unsigned 32-bit for encrypted bit

alfredh:
the ep is left-shifted by 31-bits. If the size of int
is 4 bytes then the EP bit will end up in the top bit
which is also the sign bit.

it is safer and better to use an unsigned type here

### DIFF
--- a/src/srtp/srtcp.c
+++ b/src/srtp/srtcp.c
@@ -34,7 +34,7 @@ int srtcp_encrypt(struct srtp *srtp, struct mbuf *mb)
 	struct comp *rtcp;
 	uint32_t ssrc;
 	size_t start;
-	int ep = 0;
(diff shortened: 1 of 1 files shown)


richaas:
For the record, on my linux x86_64 system, no changes to the complied binary are introduced by this PR:

$ md5sum /tmp/srtcp.o build-x86_64/srtp/srtcp.o
1108b7a14894eed123cd905dcdc45387  /tmp/srtcp.o
1108b7a14894eed123cd905dcdc45387  build-x86_64/srtp/srtcp.o
$ diff -su /tmp/srtcp.o build-x86_64/srtp/srtcp.o
Files /tmp/srtcp.o and build-x86_64/srtp/srtcp.o are identical

PR #42, opened 2017-02-20
[WIP] replacing plus delimeter

elektm93:
Implemented that the '+' symbol can be used as an 'space'. It is
replaced that commands that need a space can be used by http calls.

Baresip Example: http://0.0.0.0:8008/?/auplay+alsa,default

But I honestly doesn't now if this is according the intentions of baresip/libre.

### DIFF
--- a/src/fmt/print.c
+++ b/src/fmt/print.c
@@ -550,6 +550,24 @@ static int print_handler_dyn(const char *p, size_t size, void *arg)
 }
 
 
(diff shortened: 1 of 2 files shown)


richaas:
Hi @elektm93, thanks for your PR. We would like to keep the URL and its parameteres in its original raw format. I would recommend that you do URL unescaping inside your application.

PR #43, opened 2017-03-01
ice: make ice role type public

alfredh:
- add "ice/ICE" prefix to enum ice_role and values
- export enum ice_role to public API


alfredh:
@richaas please review and merge to master

PR #46, opened 2017-03-06
tls: add accessor to OpenSSL context (ref #35)

alfredh:
ref https://github.com/creytiv/re/pull/35

@lgrahl would this work for you?


lgrahl:
I think it would have been better to be able to set DH parameters directly in re because that is an essential part of (D)TLS which nobody should miss nowadays. So, pretty much everyone needs to implement it anyway.

As a side note: Having read through some of the code, my understanding was that re provides an OpenSSL-independent API (which is why I've made the PR in the first place). This change breaks that independence which feels a bit odd. On the other hand, it makes re extendable without any dirty hacks which is good.

While exposing the context is a good idea, I don't think it should be a replacement for the DH PR.

PR #47, opened 2017-03-16
include stdint.h to fix cross compilation on linux

ssalenik:
I'm working on a yocto recipe for libre (as well as librem and baresip) and I'm getting compilation errors when targeting linux on arm:
re_types.h:57:9: error: unknown type name 'uint32_t'

This patch fixes that... though I'm not sure if its the best way of doing it to remain compatible on other platforms.

### DIFF
--- a/include/re_types.h
+++ b/include/re_types.h
@@ -5,6 +5,7 @@
  */
 
 #include <sys/types.h>
+#include <stdint.h>
 
 #ifdef _MSC_VER
 #include <stdlib.h>


alfredh:
when compiling libre and applications using libre, you must define `HAVE_INTTYPES_H`
it should be automatically detected by re.mk if `inttypes.h` is present.

if not, you can force it:

```
$ make HAVE_INTTYPES_H=1
```

Please see here for projects that are successfully cross-compiling libre:

* https://github.com/openwrt/telephony/blob/master/libs/re/Makefile
* https://github.com/alfredh/baresip-android/blob/master/Makefile

ssalenik:
Ah, right, I had the SYSROOT incorrectly set in my initial recipe, it works now. Thanks.

PR #49, opened 2017-03-27
Handle %h prefixed (short) types

lgrahl:
I noticed that the print output when using `PRIu16` for `uint16_t` failed on OSX with a segfault. Turns out that `PRIu16 = "hu"` on this platform and re cannot handle `%h<type>`.

Note that I haven't added a new `LENMOD_SHORT` because `short` will be promoted to `int` when using vararg functions.

I have also updated `retest` (PR following).

### DIFF
--- a/src/fmt/print.c
+++ b/src/fmt/print.c
@@ -299,6 +299,11 @@ int re_vhprintf(const char *fmt, va_list ap, re_vprintf_h *vph, void *arg)
 			}
 			break;
 
+		case 'h':
+			lenmod = LENMOD_NONE;
+			fm = true;
+			break;
+
 		case 'H':
 			ph     = va_arg(ap, re_printf_h *);
 			ph_arg = va_arg(ap, void *);


alfredh:
this patch looks fine to me.

I have tested it with retest+your PR on Android NDK r13 (arm) and Debian Testing (amd64).

richaas:
IMO the modifiers %h and %hh are pretty useless, and I don't see why we should add code that do not serve any meaningful purpose. re_printf() was never supposed to be fully compliant with printf(). I suggest you just stop using %h in your code. More details [here](http://stackoverflow.com/questions/4586962/what-is-the-purpose-of-the-h-and-hh-modifiers-for-printf).

lgrahl:
@richaas I'm using the `PRI...` macros and I have no influence on the content. Meaning, I don't use `%hu` by choice. And simply ignoring `%h` instead of failing is a useful feature in that context. That's entirely independent from the usefulness of the `%h` prefix itself (and I agree with you that it's useless... but that's not the point).

alfredh:
hmm .. now I am not sure if we should merge this ..

I have always used `%u` to print `uint16_t` and it has been working fine.
I have never used the `PRIu16` macros, because I think it makes the code look ugly.

lgrahl:
Well, you can probably understand that I'm not going to discard these macros. So, for me, it's a necessity. If you don't want it, I'll have to keep a fork, redefine the macro or something else which I'd **really** want to avoid - unless you guys see another easy solution (that doesn't involve stop using these macros).

alfredh:
so ... I have done some more thinking :)

even if %h is useless I suggest that we add it (like this PR does).
it will solve Lennart's case where the application is using PRIu16 macros.

since `PRI...` is part of C99 it is probably quite likely that other applications
using libre (either now or in the future) will use a combination
of `re_printf` and `PRI...`

richaas:
I don't agree. re_printf() was never intended to support all the PRI macros, and never will. Let's agree on %h not being at all useful and forget about this.

alfredh:
okay, then we have concluded not to add it.

PR #50, opened 2017-03-27
Makefile improvements

lgrahl:
This PR provides the following improvements:
* `PREFIX` can be provided by the user
* Set `Libs.private` field in the pkg-config file
* ~~Add `install-static` to install the static library only~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,12 +34,10 @@ MODULES += odict
 MODULES += json
 
 INSTALL := install
-ifndef PREFIX
 ifeq ($(DESTDIR),)
-PREFIX  := /usr/local
+PREFIX  ?= /usr/local
 else
-PREFIX  := /usr
-endif
+PREFIX  ?= /usr
 endif
 ifeq ($(LIBDIR),)
 LIBDIR  := $(PREFIX)/lib
--- a/Makefile
+++ b/Makefile
@@ -105,16 +105,6 @@ clean:
 	@rm -rf $(SHARED) $(STATIC) libre.pc test.d test.o test $(BUILD)
 
 
-install: $(SHARED) $(STATIC) libre.pc
-	@mkdir -p $(DESTDIR)$(LIBDIR) $(DESTDIR)$(LIBDIR)/pkgconfig \
-		$(DESTDIR)$(INCDIR) $(DESTDIR)$(MKDIR)
-	$(INSTALL) -m 0644 $(shell find include -name "*.h") \
-		$(DESTDIR)$(INCDIR)
-	$(INSTALL) -m 0755 $(SHARED) $(DESTDIR)$(LIBDIR)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -112,6 +115,15 @@ install: $(SHARED) $(STATIC) libre.pc
...
 
+install-static: $(STATIC) libre.pc
+	@mkdir -p $(DESTDIR)$(LIBDIR) $(DESTDIR)$(LIBDIR)/pkgconfig \
+		$(DESTDIR)$(INCDIR) $(DESTDIR)$(MKDIR)
+	$(INSTALL) -m 0644 $(shell find include -name "*.h") \
+		$(DESTDIR)$(INCDIR)
+	$(INSTALL) -m 0755 $(STATIC) $(DESTDIR)$(LIBDIR)
+	$(INSTALL) -m 0644 libre.pc $(DESTDIR)$(LIBDIR)/pkgconfig
+	$(INSTALL) -m 0644 $(MK) $(DESTDIR)$(MKDIR)
+
```
alfredh: it has been mentioned before, that the install-static target has some duplicated code
with the `install` target.

do you think it would be possible to rewrite it a bit, so that we can have some shared code
for `install` and `install-static` ?

lgrahl: I agree that it would be great to avoid code duplication but my knowledge about Makefiles is poor. Maybe it works if `install` and `install-static` depend on another target where the shared code can be placed?

alfredh: I can try to add install-static in a separate patch.

could you please the install-static hunk from this PR?

lgrahl:
@alfredh Found a way to avoid code duplication for `install` and `install-static`. Okay now or do you want me to remove `install-static`?

alfredh:
this patch now looks fine to me.

@richaas if you agree, please review and merge to master.

Makefile
```diff
@@ -34,11 +34,13 @@ MODULES += odict
 MODULES += json
 
 INSTALL := install
+ifndef PREFIX
```
richaas: You may use PREFIX ?= below instead of ifndef

richaas:
Why do we want/need a separate make install-static target?

lgrahl:
It's convenient to ensure that only the static library can be linked. The Cmake `pkg-config` module for example is a bit quirky and only spits out library names but not their absolute paths, leading to the shared library being linked even though you've used the *static* prefixed macros. The same happens if you use the raw output of `pkg-config --static --libs libre` - you get the directory only which is error-prone when it comes to linking if both shared and static library versions are present.

richaas:
I'm ok with the PREFIX and libre.pc changes, but I'm not convinced the issue you have with linking should be solved in our Makefile. I suggest we keep install only for now, but commit the other changes.

alfredh:
the latest patch looks good to me, should be okay to merge to master now.

PR #53, opened 2017-04-03
Fixed grammar error in readme

matthew-morrison:
This is a simple edit to the readme file intending to fix a grammar error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Copyright (C) 2010 - 2017 Creytiv.com
 
 ## Building
 
-libre is using GNU makefiles, and OpenSSL development headers must be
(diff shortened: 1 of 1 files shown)


alfredh:
hi @matthew-morrison thanks for your patch

is there any reason why you have just fixed 1 line in the README ? not sure if I understand
why it is a grammatical error ..

while you are at it, could you please review the whole README file and other documents
in libre?

matthew-morrison:
Hi, I was reading the readme and that one line just stuck out at me. 

Make is not always installed on systems by default, so this just makes it more clear that you have to install it.

alfredh:
@richaas ok to merge

PR #55, opened 2017-04-16
Update for re/rem/baresip windows projects

mikhail-barg:
I'm planning to use baresip under windows, so I've updated the MSVS projects in whole baresip/rem/re combo to vs2015 (and to build at all). Also I've changed project filters to reflect the source structure.

The project is set up so that it does not require anything external, except of the following files, that should be placed in the `../misc` folder:
* unistd.h: http://stackoverflow.com/questions/341817/is-there-a-replacement-for-unistd-h-for-windows-visual-c
* getopt.h/.cpp: https://gist.github.com/ashelly/7776712

To avoid openssl dependence, I've removed `tls` module from the project.

see also https://github.com/alfredh/baresip/issues/212


src/lock/win32/lock.c
```diff
@@ -5,7 +5,9 @@
  */
 #undef  _WIN32_WINNT
 #define _WIN32_WINNT 0x0400
+#define WIN32_LEAN_AND_MEAN
 #include <windows.h>
+#undef WIN32_LEAN_AND_MEAN
```
alfredh: would it be better to move the WIN32_LEAN_AND_MEAN macro to the VS project file?

src/tcp/tcp.c
```diff
@@ -48,6 +48,10 @@
 #define SIZ_CAST
 #endif
 
+#ifdef WIN32
+#pragma comment(lib,  "ws2_32.lib") 
+#endif
+
```
alfredh: this gives a warning when using mingw32 compiler:

(ignoring #pragma comments)

alfredh:
thank you for this nice patch, it is nice to update the Visual Studio project files.

I have made a couple of inline comments.

a couple of ccheck warnings must also be fixed.

src/lock/win32/lock.c
mikhail-barg: Actually I just kept it consistent with existing files (they had WIN32_LEAN_AND_MEAN defined locally). Now I moved define to project settings. Still I'm not sure how it would affect mingw build. Maybe some makefile should be changed as well?

src/tcp/tcp.c
mikhail-barg: Moved to project file.

alfredh:
I have tested the complete PR with i686-w64-mingw32-gcc and
it compiles and runs fine.

Should be okay to merge to master now.

PR #56, opened 2017-04-26
Remove ICE Session object

alfredh:
the PR removes the ICE Session object (`struct ice`) and makes the API slightly more
user friendly and flexible.

It is related to the ICE stacks merge work, see:

https://github.com/creytiv/re/wiki/ICE-Stacks-merge

@lgrahl and @lminiero your comments are welcome.

/Alfred


lminiero:
Not sure if I was mentioned by mistake? :smile: 

I was indeed interested in having a look at the re ICE stack, but never got to it so far, so I'm afraid I wouldn't be able to provide useful feedback on the changes in the code itself. Anyway, looking at the goals in the wiki document you pointed to, the refactoring definitely seems to go in the right direction: for instance, trickle support is definitely a good thing (I wasn't aware it was only partly there so far).

A couple of notes, though:

1. When you say "gathering in app", does this mean that gathering needs to be manually started by the application, rather than started automatically as part of a different request that triggered the library? Would callbacks be involved any time a candidate is gathered, considering the gathering process might involve a server (STUN/TURN) and as such be asynchronous?
2. On the binding, it think it might make sense to let the application choose what to bind to: e.g., everything, only one interface, two out of three, etc. Not exactly sure on how this could be best formalized code-wise, but that would definitely be a useful option. Just as FYI (in case you're not aware of this already), in libnice for instance there's a method you can use to specify the addresses you'd like to be used for gathering (if you don't use it, it tries them all). If the question actually was "should we use an IP or an interface name as a constraint", I guess it doesn't really matter, and it could be both: if it's an interface name, resolve the address(es) first, and then enforce those.

Hope this helps, and sorry if I can't be of more help code-wise.

lgrahl:
@lminiero:
1. *gathering in app* means that the whole candidate gathering process is completely up to the application. There is no functionality that gathers candidates for you.
2. I guess this actually means to what address a candidate's socket is bound (see `bind`). But I'm not 100% sure.

Also, maybe the ICE stack discussion should be done somewhere else but in this issue. Maybe via gitter.im?

lgrahl:
@alfredh I'll have a look at the changes when I'm a little more awake. :)

PR #57, opened 2017-04-29
Make sampling rate for telephone events configurable

janh:
For telephone events the same timestamp base as for audio is used. Thus the correct clock rate also has to
be taken into account for calculation of durations and needs to be configurable.

### DIFF
--- a/include/re_telev.h
+++ b/include/re_telev.h
@@ -13,7 +13,7 @@ struct telev;
 
 extern const char telev_rtpfmt[];
 
-int telev_alloc(struct telev **tp, uint32_t ptime, uint32_t srate);
+int telev_alloc(struct telev **tp, uint32_t ptime);
 int telev_set_srate(struct telev *tel, uint32_t srate);
 int telev_send(struct telev *tel, int event, bool end);
 int telev_recv(struct telev *tel, struct mbuf *mb, int *event, bool *end);
--- a/include/re_telev.h
+++ b/include/re_telev.h
@@ -13,7 +13,8 @@ struct telev;
 
 extern const char telev_rtpfmt[];
 
-int telev_alloc(struct telev **tp, uint32_t ptime);
+int telev_alloc(struct telev **tp, uint32_t ptime, uint32_t srate);
+int telev_set_srate(struct telev *tel, uint32_t srate);
 int telev_send(struct telev *tel, int event, bool end);
 int telev_recv(struct telev *tel, struct mbuf *mb, int *event, bool *end);
 int telev_poll(struct telev *tel, bool *marker, struct mbuf *mb);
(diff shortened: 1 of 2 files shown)


src/telev/telev.c
```diff
@@ -177,6 +180,24 @@ int telev_alloc(struct telev **tp, uint32_t ptime)
 
 
 /**
+ * Sets the sampling rate
+ *
+ * @param srate Sampling rate in [Hz]
+ *
```
alfredh: the `tel` param should be added to the list

src/telev/telev.c
```diff
@@ -177,6 +180,24 @@ int telev_alloc(struct telev **tp, uint32_t ptime)
...
+ * @return 0 if success, otherwise errorcode
+ */
+int telev_set_srate(struct telev *tel, uint32_t srate)
+{
+	if (!tel)
+		return EINVAL;
+
```
alfredh: if the `srate` value is zero it should return EINVAL

src/telev/telev.c
```diff
@@ -275,15 +296,15 @@ int telev_poll(struct telev *tel, bool *marker, struct mbuf *mb)
...
+		tel->dur   = tel->ptime * (tel->srate / 1000);
 		tel->state = SENDING;
 		tel->txc   = 1;
 
 		err = payload_encode(mb, tel->event, false, tel->dur);
 		break;
 
 	case SENDING:
-		tel->dur += tel->ptime * 8;
+		tel->dur += tel->ptime * (tel->srate / 1000);
```
alfredh: this could be optimized.

if the value of ptime * srate / 1000 is calculated in the alloc function and saved,
it can be used instead. Thus avoiding a multiplication and division for every packet sent.

include/re_telev.h
```diff
@@ -13,7 +13,8 @@ struct telev;
 
 extern const char telev_rtpfmt[];
 
-int telev_alloc(struct telev **tp, uint32_t ptime);
+int telev_alloc(struct telev **tp, uint32_t ptime, uint32_t srate);
```
alfredh: this will break the API, are we sure we want to do this ?

alfredh:
hi @janh 

many thanks for your nice patch.

from RFC 4733:

```
   The default clock frequency
   is 8000 Hz, but the clock frequency can be redefined when assigning
   the dynamic payload type.
```

perhaps we should also use 8000 Hz as the default srate, and let the
application change it if needed?

not sure if we should break the API with this. Additional comments inline.



/Alfred

janh:
I removed the srate parameter from telev_alloc and also added changes for your other comments.

alfredh:
this patch now looks good to me.

@richaas if you agree with the patch, should we merge it to master?

PR #60, opened 2017-05-23
rtp: add extension bit to the api

alfredh:
this PR adds "bool ext" to the `rtp_send` function

currently the EXTENSION bit is always set to false in the RTP header.
setting the extension bit per packet is needed for e.g. RFC 5285

NOTE! This will break the api.

if breaking the api is not wanted, we could instead add a new function
for sending with EXT=true:

```
rtp_send_ext();
```


richaas:
IMO we should try to avoid breaking the API.

alfredh:
I added new function instead:

```
rtp_send_ext()
```

richaas:
Hmm, thinking more about this I suggest we go for your first solution. Adding new functions like this will quickly lead to a messy API. If you revert your second commit, I'll merge the original version. Sorry for the back-and-forth.

alfredh:
thanks, I did instead a git cherry-pick of the first commit:

https://github.com/creytiv/re/commit/920a41d3196445842568a455ab6e8c81723cb29e

PR #62, opened 2017-06-06
Pack struct of mqueue messages.

fancycode:
Without the change, the `data` field is padded to the next 8 bytes boundary on 64 bit systems, resulting in `sizeof(struct msg)` to be 24 bytes and valgrind reporting uninitialized bytes:
```
==22350== Syscall param write(buf) points to uninitialised byte(s)
==22350==    at 0x5CB734D: ??? (syscall-template.S:81)
==22350==    by 0x948EFE: pipe_write (mqueue.h:21)
==22350==    by 0x949157: mqueue_push (mqueue.c:152)
```

After reordering the fields, the struct size is 16 bytes and no padding is required.

### DIFF
--- a/src/mqueue/mqueue.c
+++ b/src/mqueue/mqueue.c
@@ -36,9 +36,9 @@ struct mqueue {
 };
 
 struct msg {
-	int id;
 	void *data;
 	uint32_t magic;
+	int id;
 };
 
 


fancycode:
Is there anything you want to have changed here? It would be nice to be able to use unmodified upstream sources without having to apply a patch locally for running in valgrind.

alfredh:
many thanks for your patch

the memory layout is something like this:

```
.----------------.
|XXXXXXXXpppppppp|
|XXXXXXXXXXXXXXXX|
|XXXXXXXXpppppppp|
'----------------'
```
where X is used data and p is padding.

valgrind is complaining because the padding is uninitialized, however
the padding bytes are not used in the code. So the code works just fine
as it is.

another solution could be to do a `memset` on the struct.

I support this change, because it reduces the size of the struct (mqueue overhead).


/Alfred

fancycode:
Yeah, I know about padding and valgrind, that's why I created the PR :smiley: 

Using a `memset` (or the struct zero-initializer) is not necessary and might cause additional (although very little) runtime overhead.

PR #65, opened 2017-06-24
optimize rand_str() and rand_char()

richaas:
Increase performance and robustness of rand_str() and rand_char().

@alfredh please review and merge

### DIFF
--- a/src/sys/rand.c
+++ b/src/sys/rand.c
@@ -3,7 +3,6 @@
  *
  * Copyright (C) 2010 Creytiv.com
  */
-#include <ctype.h>
 #include <stdlib.h>
 #ifdef USE_OPENSSL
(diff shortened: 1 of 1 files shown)


alfredh:
This is a very nice improvement.

I tested the performance of `rand_str` on an Android (arm) device (average of 3 runs):

master:  4250 ms
rand_str:  122 ms

code:
```
void test(void)
{
	uint64_t x,y;
	char str[1024];
	size_t i;

	x = tmr_jiffies();

	for (i=0; i<1000; i++) {
		rand_str(str, sizeof(str));
	}

	y = tmr_jiffies();

	re_printf("time: %d ms\n", (int)(y-x));
}
```

so the new code is considerably faster.

/Alfred

PR #70, opened 2017-08-01
Fix compilation on OpenRISC when using OR1K.

fancycode:
OR1K (http://opencores.org/or1k/) doesn't support "res_nclose" similar to OpenBSD.


alfredh:
this patch looks fine to me.

please note that this API is only used to fetch the IP-address of the DNS-servers
from the system, in addition to /etc/resolv.conf

it can be disabled at build time with:

```
make HAVE_RESOLV=
```

alfredh:
The openrisc platform is not supported.

PR #71, opened 2017-08-02
Add function to get the remote address of the selected candidate pair

fancycode:
This matches the existing `icem_selected_laddr`.

PR #72, opened 2017-08-04
ice: Added functions to get selected candidates.

fancycode:
Also added a function to get the type of a candidate. This can be used when
sending data to handle relay candidates, where data must be sent through the
TURN client.

This is a more flexible variant of what I did in #71.

### DIFF
--- a/include/re_ice.h
+++ b/include/re_ice.h
@@ -100,6 +100,10 @@ struct list *icem_checkl(const struct icem *icem);
 struct list *icem_validl(const struct icem *icem);
 const struct sa *icem_cand_default(struct icem *icem, unsigned compid);
 const struct sa *icem_selected_laddr(const struct icem *icem, unsigned compid);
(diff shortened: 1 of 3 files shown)


alfredh:
many thanks for this PR, it looks good.

@richaas please review and merge.

can you also share your thoughts on the topic of merging
the 2 ICE-stacks:

https://github.com/creytiv/re/wiki/ICE-Stacks-merge

I am quite interested in feedback regarding various parts of the API,
i.e. which parts are good and which parts are not so good.

richaas:
I would recommend that struct ice_cand is forward declared before referencing it in function declarations.

fancycode:
> I would recommend that struct ice_cand is forward declared before referencing it in function declarations.

Thanks for the heads-up, changed.

alfredh:
many thanks. okay to merge now?

PR #79, opened 2017-09-22
Update ar flags; use deterministic mode

richaas:
In Debian 9 (stretch), the ar program in the binutils package is now configured with
`--enable-deterministic-archives`. This make deterministic default mode, opposite of previous behavior. When running ar on Debian 9 we then get the following warning:

```
  CC      build-x86_64/json/decode_odict.o
  CC      build-x86_64/json/encode.o
  LD      libre.so
  AR      libre.a
ar: `u' modifier ignored since `D' is the default (see `U')
```

This PR make ar always use deterministic mode (on linux). This change makes the archives reproducible; equal input will always result in binary equal archive. The alternative would be to add the `U` option and thus keep the old, non-deterministic behavior.

@alfredh, please review and commit.

### DIFF
--- a/mk/re.mk
+++ b/mk/re.mk
@@ -237,7 +237,7 @@ ifeq ($(OS),linux)
 	MOD_LFLAGS	+=
 	APP_LFLAGS	+= -rdynamic
 	AR		:= ar
-	AFLAGS		:= cru
+	AFLAGS		:= crD
 endif
 ifeq ($(OS),gnu)
 	CFLAGS		+= -fPIC -DGNU


alfredh:
this one looks good to go, I have tested the patch on a couple
of Linux distributions:

```
AR flags for Linux:
------------------
Android r14	gcc 4.9.x	ar 2.25.51	OK
Arch Linux:	gcc 7.2.0	ar 2.29		OK
Debian 7	gcc 4.7.2	ar 2.22		OK
Debian 8	gcc 4.9.2	ar 2.25 	OK
Fedora 24	gcc 6.3.1	ar 2.26.1	OK
Ubuntu 14.04	gcc 4.8.4	ar 2.24		OK
Ubuntu 16.04	gcc 5.4.0	ar 2.26		OK
```


the test was to build libre and retest, and execute retest -r.

PR #83, opened 2017-10-05
sip: handle SIPS URI in Record-Route

glenvt18:
If a route starts with sips:, send using TLS transport.

Some proxies insert 

`Record-Route: <sips:ADDR:PORT;..>` 

instead of 

`Record-Route: <sip:ADDR:PORT;transport=tls;...>`

In such cases the message (ACK) is not sent and the session is aborted (ENOSYS).

An example (SIP messages are dumped from the source, this PR applied):
https://pastebin.com/wkppt8h8

More info:
https://tools.ietf.org/html/rfc5630#section-3.1.3

### DIFF
--- a/src/sip/request.c
+++ b/src/sip/request.c
@@ -602,11 +602,14 @@ int sip_request(struct sip_request **reqp, struct sip *sip, bool stateful,
 	struct sa dst;
 	struct pl pl;
 	int err;
+	bool sips_uri = false;
 
 	if (!sip || !met || !uri || !route || !mb)
 		return EINVAL;
 
-	if (pl_strcasecmp(&route->scheme, "sip"))
+	sips_uri = (pl_strcasecmp(&route->scheme, "sips") == 0);
+
+	if (pl_strcasecmp(&route->scheme, "sip") && !sips_uri)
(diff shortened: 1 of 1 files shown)


glenvt18:
As it looks like sips scheme is not going to be supported by libre I'm closing this PR.

lgrahl:
Not that I'm active in the SIP area but any particular reason why you're closing this PR?

glenvt18:
@lgrahl https://github.com/alfredh/baresip/issues/303. This PR provides rather a hack/workaround which actually downgrades sips to sip+tls. Though, it can be used to solve linphone.org issue if you know what you're doing - baresip doesn't support sips calls anyway. Perhaps limiting this to routes only (while making a sip, not sips, call) might be reasonable. Full sips support will require much more effort, and the current policy is to stay away from sips altogether.

BTW. I'm not a SIP expert too, but I wonder why the linphone proxy adds that sips:... route. Proxy->PBX hop is insecure, and, if you call `sips:thetestcall@sip.linphone.org`, they answer `480 Temporarily Unavailable` with `SIPS not allowed`.

PR #84, opened 2017-10-13
sip: configurable timer values

alfredh:
Add support for setting SIP Timers T1 T2 T4

use `sip_conf_set` to set the config
use `sip_conf` to get the config

### DIFF
--- a/include/re_sip.h
+++ b/include/re_sip.h
@@ -221,6 +221,13 @@ struct sip_contact {
 	enum sip_transp tp;
 };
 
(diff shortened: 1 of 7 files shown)


alfredh:
I have added inline functions to access timer values (`sip_t1`)

we could make them uppercase (`SIP_T1`) but then we have to rename
the default timer values.

alfredh:
requirements are unclear, closing PR for now and continue hacking in branch

PR #89, opened 2017-10-30
tls: add dtls_recv_packet()

alfredh:
The current DTLS module only supports receiving packets on the UDP socket.

There is a usecase where an incoming DTLS packet will arrive on a different type
of socket. If TURN over TCP/TLS is used for the media transport, the DTLS packets
will be sent/recv via a TCP socket.

If ICE is used as media transport, with both UDP-based and TCP-based candidates,
the DTLS module must be able to receive packets on both of them.

@richaas please review and merge to master

### DIFF
--- a/src/tls/openssl/tls_udp.c
+++ b/src/tls/openssl/tls_udp.c
@@ -829,6 +829,20 @@ void dtls_set_mtu(struct dtls_sock *sock, size_t mtu)
 }
 
 
+void dtls_recv_packet(struct dtls_sock *sock, const struct sa *src,
+		      struct mbuf *mb)
+{
+	struct sa addr;
+
+	if (!sock || !src || !mb)
+		return;
+
+	addr = *src;
+
+	recv_handler(&addr, mb, sock);
+}
+
+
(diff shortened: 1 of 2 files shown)


src/tls/openssl/tls_udp.c
```diff
@@ -829,6 +829,20 @@ void dtls_set_mtu(struct dtls_sock *sock, size_t mtu)
...
 
+void dtls_recv_packet(struct dtls_sock *sock, const struct sa *src,
+		      struct mbuf *mb)
+{
+	struct sa addr;
+
+	if (!sock || !src || !mb)
+		return;
+
+	addr = *src;
```
lgrahl: I'm curious: Why is the copy needed?

alfredh: the input address is const, but the recv_handler is non-const.

we cannot simply cast away the const-ness, since the recv_handler function
might alter the address.

lgrahl: Ah, missed that it's const in the function's signature. Thanks for the explanation. :+1:

PR #90, opened 2017-11-01
http: added support for chunked transfer encoding

richaas:
This PR adds support for chunked transfer encoding in the HTTP client. This transfer encoding is quite common, and support for this is mandatory in [HTTP/1.1](https://tools.ietf.org/html/rfc7230#section-4.1). @alfredh please test, review and merge to master.

### DIFF
--- a/include/re_http.h
+++ b/include/re_http.h
@@ -82,7 +82,8 @@ struct http_msg {
 	struct pl reason;      /**< Response Reason phrase                 */
 	struct list hdrl;      /**< List of HTTP headers (struct http_hdr) */
 	struct msg_ctype ctyp; /**< Content-type                           */
-	struct mbuf *mb;       /**< Buffer containing the HTTP message     */
+	struct mbuf *_mb;      /**< Buffer containing the HTTP message     */
+	struct mbuf *mb;       /**< Buffer containing the HTTP body        */
 	uint32_t clen;         /**< Content length                         */
 };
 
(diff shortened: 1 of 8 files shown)


include/re_http.h
```diff
@@ -114,16 +115,20 @@ int  http_msg_print(struct re_printf *pf, const struct http_msg *msg);
...
 struct dnsc;
+struct tcp_conn;
+struct tls_conn;
 
 typedef void (http_resp_h)(int err, const struct http_msg *msg, void *arg);
-typedef void (http_data_h)(struct mbuf *mb, void *arg);
+typedef int  (http_data_h)(const uint8_t *buf, size_t size,
+			   const struct http_msg *msg, void *arg);
+typedef void (http_conn_h)(struct tcp_conn *tc, struct tls_conn *sc,
+			   void *arg);
```
alfredh: perhaps add a doxygen comment to the prototypes, documenting the semantics of the
callback handler and the arguments.

src/websock/websock.c
```diff
@@ -412,13 +412,8 @@ static void http_resp_handler(int err, const struct http_msg *msg, void *arg)
 	/* here we are ok */
 
 	conn->state = OPEN;
-	conn->tc = mem_ref(http_req_tcp(conn->req));
-	conn->sc = mem_ref(http_req_tls(conn->req));
 	(void)tcp_conn_peer_get(conn->tc, &conn->peer);
```
alfredh: looks like `conn->tc` is accessed here, is the conn->tc NULL?

richaas: No, conn->tc is set in the conn_handler(), which is called just before the response handler.

include/re_http.h
richaas: That make sense, but as they were not commented initially, let's do that in a separate PR.

richaas: The connection handler was added to provide better support for applications that needs
to continue communication on the existing TCP connection after the HTTP response is received. Examples include web-socket clients and streaming clients. If this handler is set, the HTTP client will not reuse this connection in new HTTP requests.

richaas: [HTTP/2](https://tools.ietf.org/html/rfc7540#section-3.2) is another example where this is useful.

alfredh:
updated testcase in retest to cover chunked response:

https://github.com/creytiv/retest/commit/f53e2ebdde48a1c0adcbcb610aa39e83f1ffd8c1

PR #97, opened 2017-11-28
Fix network on Windows

ftk:
On my machine socket function returns fd=~500-600, which triggers an error since it is more than FD_SETSIZE(64).
Redefining FD_SETSIZE to a larger value fixes this error, but select wont trigger.

In this PR I replaced select with WSAPoll and increased FD_SETSIZE. WSAPoll is not available on WinXP.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ libre.a
 libre.dylib
 libre.pc
 libre.so
+libre.dll
--- a/mk/re.mk
+++ b/mk/re.mk
@@ -324,7 +324,7 @@ ifeq ($(OS),openbsd)
 	HAVE_ARC4RANDOM	:= 1
 endif
 ifeq ($(OS),win32)
-	CFLAGS		+= -DWIN32 -D_WIN32_WINNT=0x0501 -D__ssize_t_defined
+	CFLAGS		+= -DWIN32 -D_WIN32_WINNT=0x0601 -D__ssize_t_defined
 	LIBS		+= -lwsock32 -lws2_32 -liphlpapi
 	LFLAGS		+=
 	SH_LFLAGS	+= -shared
@@ -548,6 +548,8 @@ endif
 ifeq ($(OS),win32)
 CFLAGS  += -DHAVE_SELECT
 CFLAGS  += -DHAVE_IO_H
+CFLAGS  += -DHAVE_INET_NTOP -DHAVE_INET_PTON
+CFLAGS  += -DFD_SETSIZE=1024 -DHAVE_POLL
 else
 HAVE_SYSLOG  := $(shell [ -f $(SYSROOT)/include/syslog.h ] && echo "1")
 HAVE_DLFCN_H := $(shell [ -f $(SYSROOT)/include/dlfcn.h ] && echo "1")
--- a/src/main/main.c
+++ b/src/main/main.c
@@ -3,6 +3,16 @@
  *
  * Copyright (C) 2010 Creytiv.com
  */
+#ifdef HAVE_POLL
+#ifdef WIN32
+#define _WIN32_WINNT 0x601
+#define WINVER 0x601
+#include <winsock2.h>
+#else
+#include <poll.h>
+#endif
+#endif
+
 #ifdef HAVE_SYS_TIME_H
 #include <sys/time.h>
 #endif
@@ -21,9 +31,7 @@
 #ifdef HAVE_SELECT_H
 #include <sys/select.h>
 #endif
-#ifdef HAVE_POLL
-#include <poll.h>
-#endif
+
 #ifdef HAVE_EPOLL
 #include <sys/epoll.h>
 #endif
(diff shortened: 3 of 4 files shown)


src/net/win32/wif.c
```diff
@@ -24,7 +24,7 @@
  */
 static int if_list_gaa(net_ifaddr_h *ifh, void *arg)
 {
-	IP_ADAPTER_ADDRESSES addrv[16], *cur;
+	IP_ADAPTER_ADDRESSES addrv[64], *cur;
```
alfredh: could you please create a separate PR for this change?

alfredh:
I am getting warnings with mingw32 from Macports:

```
In file included from include/re_sa.h:7:0,
                 from src/sa/ntop.c:29:
/opt/local/i686-w64-mingw32/include/winsock2.h:15:2: warning: #warning Please include winsock2.h before windows.h [-Wcpp]
 #warning Please include winsock2.h before windows.h
  ^~~~~~~
src/sa/ntop.c: In function ‘net_inet_ntop’:
src/sa/ntop.c:208:22: warning: passing argument 2 of ‘inet_ntop’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
   inet_ntop(AF_INET, &sa->u.in.sin_addr, buf, size);
                      ^
In file included from include/re_sa.h:8:0,
                 from src/sa/ntop.c:29:
/opt/local/i686-w64-mingw32/include/ws2tcpip.h:447:35: note: expected ‘PVOID {aka void *}’ but argument is of type ‘const struct in_addr *’
 WINSOCK_API_LINKAGE LPCSTR WSAAPI InetNtopA(INT Family, PVOID pAddr, LPSTR pStringBuf, size_t StringBufSize);
                                   ^
src/sa/ntop.c:213:23: warning: passing argument 2 of ‘inet_ntop’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
   inet_ntop(AF_INET6, &sa->u.in6.sin6_addr, buf, size);
                       ^
In file included from include/re_sa.h:8:0,
                 from src/sa/ntop.c:29:
/opt/local/i686-w64-mingw32/include/ws2tcpip.h:447:35: note: expected ‘PVOID {aka void *}’ but argument is of type ‘const struct in6_addr *’
 WINSOCK_API_LINKAGE LPCSTR WSAAPI InetNtopA(INT Family, PVOID pAddr, LPSTR pStringBuf, size_t StringBufSize);
                                   ^
```

.gitignore
```diff
@@ -9,3 +9,4 @@ libre.a
 libre.dylib
 libre.pc
 libre.so
+libre.dll
```
alfredh: please create a separate PR for this change..

mk/re.mk
```diff
@@ -324,15 +319,14 @@ ifeq ($(OS),openbsd)
 	HAVE_ARC4RANDOM	:= 1
 endif
 ifeq ($(OS),win32)
-	CFLAGS		+= -DWIN32 -D_WIN32_WINNT=0x0501 -D__ssize_t_defined
+	CFLAGS		+= -DWIN32 -D_WIN32_WINNT=0x0601 -D__ssize_t_defined
```
alfredh: what is the reason for this change?

perhaps we should document these magic values.

alfredh:
many thanks for your patch.

In general we should find a solution that works both on Windows XP and Vista.

ftk:
-D_WIN32_WINNT=0x0501 means to compile for Windows NT 5.1 (XP)
WSAPoll only gets defined if _WIN32_WINNT >= 0x0600
To maintain compatibility with WinXP you would have to copy definitions from winsock2.h, then at runtime detect windows version and either use select or load WSAPoll function pointer from some DLL.
I rather might add support for IO completion ports in the future, which also should be faster than iterating 600+ fds every time.
_WIN32_WINNT >= 0x0600 also defines inet_ntop and inet_pton which have slightly different signature, therefore the warnings. It is possible to add #undef inet_ntop  #undef inet_pton and use library's code instead.

src/main/main.c
```diff
@@ -101,8 +107,12 @@ struct re {
 	struct list tmrl;            /**< List of timers                    */
 
 #ifdef HAVE_POLL
+#ifdef WIN32
+	WSAPOLLFD *fds;
+#else
```
alfredh: perhaps we should add a `HAVE_WSAPOLL` flag ?

it would be nice if the same codebase can be used to build for different
platforms, by setting flags.

alfredh:
perhaps we should also consider this:

https://daniel.haxx.se/blog/2012/10/10/wsapoll-is-broken/

alfredh:
We should not add support for WSA poll as it is broken. It was removed from CURL.

I have added IOCP to the roadmap/wishlist:

https://github.com/creytiv/re/wiki/Roadmap

if you still want to contribute, please break your PR up into smaller patches,
so they can be reviewed and tested.

Alfred

PR #98, opened 2017-11-29
Fix ERROR_BUFFER_OVERFLOW when calling GetAdaptersAddresses 

ftk:
https://msdn.microsoft.com/en-us/library/windows/desktop/aa365915(v=vs.85).aspx

### DIFF
--- a/src/net/win32/wif.c
+++ b/src/net/win32/wif.c
@@ -24,7 +24,7 @@
  */
 static int if_list_gaa(net_ifaddr_h *ifh, void *arg)
 {
-	IP_ADAPTER_ADDRESSES addrv[16], *cur;
+	IP_ADAPTER_ADDRESSES addrv[64], *cur;
 	ULONG ret, len = sizeof(addrv);
 	const ULONG flags = GAA_FLAG_SKIP_ANYCAST | GAA_FLAG_SKIP_MULTICAST;
 	HANDLE hLib;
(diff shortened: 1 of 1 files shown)


src/net/win32/wif.c
```diff
@@ -46,14 +47,24 @@ static int if_list_gaa(net_ifaddr_h *ifh, void *arg)
 		goto out;
 	}
 
-	ret = (*u.gaa)(AF_UNSPEC, flags, NULL, addrv, &len);
+	/* try with stack-allocated addrv at first */
+	paddrv = addrv;
+	addrv_allocated = false;
+retry:
```
alfredh: we should avoid a retry loop here, since there is a chance that the code will be stuck
in the loop forever.

alfredh:
many thanks for your patch

I think we should find a solution that uses the same code path for all cases.

with your patch, the error case would only be done in certain cases.
perhaps it is better to just increase the buffer to 64 ?

another solution is to always query the required buffer size:

```
	/* fetch the required buffer size */
	ret = (*u.gaa)(AF_UNSPEC, flags, NULL, NULL, &len);

	addrv = mem_zalloc(len, NULL);

	ret = (*u.gaa)(AF_UNSPEC, flags, NULL, addrv, &len);

```

alfredh:
I have made an alternative PR here:

https://github.com/creytiv/re/pull/100

Alfred

PR #99, opened 2017-11-30
tmr: use clock_gettime() to avoid system clock drift

alfredh:
the current timer system is subject to adjusting the system clock.

1. for a pending timer, if the system clock is adjusted towards the future,
  the timer will trigger before it is supposed to trigger.

2. for a pending timer, if the system clock is adjusted towards the past,
  the timer will trigger later.

one solution to avoid this is to use `clock_gettime` for systems that support it:

```
 CLOCK_MONOTONIC

clock that increments monotonically, tracking the time since an arbitrary
point, and will continue to increment while the system is asleep.
```

https://linux.die.net/man/3/clock_gettime

Testing:

- [x] Debian Testing
- [x] OSX 10.12
- [x] Android NDK r14
- [x] OpenBSD 6.2

### DIFF
--- a/src/tmr/tmr.c
+++ b/src/tmr/tmr.c
@@ -3,6 +3,8 @@
  *
  * Copyright (C) 2010 Creytiv.com
  */
+
+#define _DEFAULT_SOURCE 1
 #include <string.h>
 #ifdef HAVE_SYS_TIME_H
 #include <sys/time.h>
@@ -132,6 +134,26 @@ uint64_t tmr_jiffies(void)
 	li.LowPart = ft.dwLowDateTime;
 	li.HighPart = ft.dwHighDateTime;
 	jfs = li.QuadPart/10/1000;
+#elif HAVE_CLOCK_GETTIME
+
+	struct timespec now;
+	clockid_t clock_id;
+
+#if defined (CLOCK_MONOTONIC_RAW_APPROX)
+	clock_id = CLOCK_MONOTONIC_RAW_APPROX;
+#elif defined (CLOCK_BOOTTIME)
+	clock_id = CLOCK_BOOTTIME;
+#else
+	clock_id = CLOCK_MONOTONIC;
+#endif
+
+	if (0 != clock_gettime(clock_id, &now)) {
+		DEBUG_WARNING("jiffies: clock_gettime() failed (%m)\n", errno);
+		return 0;
+	}
+
+	jfs  = (long)now.tv_sec * (uint64_t)1000;
+	jfs += now.tv_nsec / (uint64_t)1000000;
 #else
 	struct timeval now;
 
--- a/src/tmr/tmr.c
+++ b/src/tmr/tmr.c
@@ -139,9 +139,7 @@ uint64_t tmr_jiffies(void)
 	struct timespec now;
 	clockid_t clock_id;
 
-#if defined (CLOCK_MONOTONIC_RAW_APPROX)
-	clock_id = CLOCK_MONOTONIC_RAW_APPROX;
-#elif defined (CLOCK_BOOTTIME)
+#if defined (CLOCK_BOOTTIME)
 	clock_id = CLOCK_BOOTTIME;
 #else
 	clock_id = CLOCK_MONOTONIC;


richaas:
CLOCK_MONOTONIC is affected by the incremental adjustments performed by adjtime(3) and NTP.
On Linux, CLOCK_MONOTONIC is not incremented during suspend.

alfredh:
do you think we should add support for clock_gettime ?

There are some interesting articles here:

http://btorpey.github.io/blog/2014/02/18/clock-sources-in-linux/

https://blog.habets.se/2010/09/gettimeofday-should-never-be-used-to-measure-time.html

since the value returned by `tmr_jiffies` is in the units of "milliseconds" it is perhaps
better for us to use a coarse (not high precision) timer that is cheaper to fetch.

from the manual page of clock_gettime on OSX:

```
  CLOCK_MONOTONIC    clock that increments monotonically, tracking the time since an
                        arbitrary point, and will continue to increment while the system is
                        asleep.

     CLOCK_MONOTONIC_RAW
                        clock that increments monotonically, tracking the time since an
                        arbitrary point like CLOCK_MONOTONIC.  However, this clock is unaf-
                        fected by frequency or time adjustments.  It should not be compared
                        to other system time sources.

     CLOCK_MONOTONIC_RAW_APPROX
                        like CLOCK_MONOTONIC_RAW, but reads a value cached by the system at
                        context switch.  This can be read faster, but at a loss of accuracy
                        as it may return values that are milliseconds old.
```

the original issue came up on a device that started with time 1. Jan 1970
and was later set to the correct time by NTP. In this case the time delta was
so large that it made all the re timers expire immediately.

richaas:
I support the idea of improving tmr_jiffes(), and make it resilient agains time adjustments.
However, the problem with CLOCK_MONOTONIC is that it is not incremented during system
suspend on Linux (but is on OS X). Additionally, CLOCK_MONOTONIC is influenced by NTP
adjustments.

fancycode:
You could use `CLOCK_BOOTTIME` on Linux to include time while the system is suspended:
```
       CLOCK_BOOTTIME (since Linux 2.6.39; Linux-specific)
              Identical  to  CLOCK_MONOTONIC,  except  it  also  includes any time that the system is suspended.  This allows
              applications to get a  suspend-aware  monotonic  clock  without  having  to  deal  with  the  complications  of
              CLOCK_REALTIME, which may have discontinuities if the time is changed using settimeofday(2).
```

alfredh:
hi @fancycode many thanks for your suggestion.

I updated the patch with new clocks for Linux and OSX.

richaas:
I suggest we drop CLOCK_MONOTONIC_RAW_APPROX, and use either CLOCK_BOOTTIME (on linux) or otherwise CLOCK_MONOTONIC. We also need to test that this works on all supported platforms, and that the timer is suspend-aware on all of those. On potential issue is that older linux kernels (prior to 2.6.39) will no longer have suspend-aware timers, as they currently have.

alfredh:
@fancycode do you have the chance to test this patch on your system(s) ?

the code must be compiled with

```
$ make EXTRA_CFLAGS=-DHAVE_CLOCK_GETTIME
```


Alfred

alfredh:
I think that we are not really sure how to solve this yet.

in order to do some more experimentation and testing I will add a function
to baresip that measures time (in microseconds) using the clock_gettime.

fwjavox:
May I ask why you stick with gettimeofday instead of clock_gettime?
As @richaas pointed out, some of the clock_gettime clocks are affected by NTP adjustments, but that only applies to frequency adjustment. There won't be any jumps in the monotonic clocks (except suspends for CLOCK_BOOTTIME).
From what I understand, the only downside of CLOCK_BOOTTIME is lacking support in ancient kernels.
The downsides of gettimeofday for measuring time differences are pointed out in the link @alfredh posted above.
I wrote the patch to fix problems we had with systems that were powered off for a few days and resynced their clock using NTP which lead to timers being late.

> One potential issue is that older linux kernels (prior to 2.6.39) will no longer have suspend-aware timers, as they currently have.

This could be avoided using preprocessor defines, I guess.

alfredh:
the main problem with this patch is that it has to be tested extremely well
on a large number of devices/platforms and by many people.

I dont want to apply this patch now and find bugs after 6 months.

PR #101, opened 2017-12-13
Fix build with musl libc

strfry:
The other day i was trying to build re/baresip on Alpine Linux, which uses musl libc instead of glibc.

With a few small changes it builds just fine.

My only nitpick is one spot, where i had to work around the fact that musl-libc doesn't offer the reentrant res_ninit() API, and the code path for OpenBSD has to be taken instead.
Since musl considers #ifdef testing for libc variants evil, there is no `__MUSL__` macro to test against. I worked around it by testing for linux and absence of other libc's known to me, which is probably worse than what the musl devs intended to avoid.
I'm open to suggestions how to improve that.

### DIFF
--- a/src/dns/res.c
+++ b/src/dns/res.c
@@ -18,6 +18,12 @@
 #include <re_dns.h>
 #include "dns.h"
 
+#if defined(__linux__) && !defined(__GLIBC__) &&	\
+	!defined(__UCLIBC__) &&	!defined(__BIONIC__)
+/* Another popular libc for Linux offers OpenBSD's res_init() API */
+#define OPENBSD OPENBSD
+#endif
+
 
 int get_resolv_dns(char *domain, size_t dsize, struct sa *nsv, uint32_t *n)
 {
--- a/src/net/ifaddrs.c
+++ b/src/net/ifaddrs.c
@@ -6,6 +6,7 @@
 #include <unistd.h>
 #include <sys/socket.h>
 #define __USE_MISC 1   /**< Use MISC code */
+#define _BSD_SOURCE
 #include <net/if.h>
 #include <ifaddrs.h>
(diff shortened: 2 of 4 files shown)


src/dns/res.c
```diff
@@ -18,6 +18,12 @@
 #include <re_dns.h>
 #include "dns.h"
 
+#if defined(__linux__) && !defined(__GLIBC__) &&	\
+	!defined(__UCLIBC__) &&	!defined(__BIONIC__)
+/* Another popular libc for Linux offers OpenBSD's res_init() API */
+#define OPENBSD OPENBSD
+#endif
+
```
alfredh: this logic is not robust. what if there is a new libc project that is called e.g. MUSL_NG ?

you can work around this by building with the `HAVE_RESOLV` flag unset.

src/net/ifaddrs.c
```diff
@@ -6,6 +6,7 @@
 #include <unistd.h>
 #include <sys/socket.h>
 #define __USE_MISC 1   /**< Use MISC code */
+#define _BSD_SOURCE
```
alfredh: is it possible to use `_DEFAULT_SOURCE` instead ?

I think we are using that one in most places now

alfredh:
Hi,

many thanks for your patch. unfortunately we do not support musl.

for recent versions of glibc, we added a define of _DEFAULT_SOURCE to
some .c files it seemed to work better than _BSD_SOURCE ..

src/net/ifaddrs.c
strfry: Unfortunately, musl uses 
`#if defined(_GNU_SOURCE) || defined(_BSD_SOURCE)`
to guard definitions of IFF_UP et al, so this is not possible.

PR #102, opened 2017-12-18
mem: add secure memory functions

alfredh:
the use case for these functions is secure code like
SRTP and TLS.

memcmp() should not be used when comparing secret data
such as shared keys, because the content may affect the timing.
instead the compare function should always take the same
to execute, independently of the content.

### DIFF
--- a/src/mem/secure.c
+++ b/src/mem/secure.c
@@ -27,7 +27,7 @@ int mem_seccmp(const volatile uint8_t *volatile s1,
 {
 	uint8_t val = 0;
 
-	if (!s1 || !s2 || !n)
+	if (!s1 || !s2)
 		return -1;
 
 	while (n--)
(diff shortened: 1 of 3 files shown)


alfredh:
I have updated the patch with only function:

mem_seccmp

src/mem/secure.c
```diff
@@ -0,0 +1,37 @@
...
+ *         0 if both byte strings matching
+ *         a positive number if not matching
+ */
+int mem_seccmp(const volatile uint8_t *volatile s1,
+	       const volatile uint8_t *volatile s2,
+	       size_t n)
+{
+	uint8_t val = 0;
+
+	if (!s1 || !s2 || !n)
```
richaas: Zero length should return 0 (skip !n check)

PR #103, opened 2018-01-05
sip: add To name to sip_dialog_alloc

alfredh:
The usecase here is to be able to specify the To-header Display name.

Example message from SIP RFC:

```
       REGISTER sip:registrar.biloxi.com SIP/2.0
       Via: SIP/2.0/UDP bobspc.biloxi.com:5060;branch=z9hG4bKnashds7
       Max-Forwards: 70
       To: Bob <sip:bob@biloxi.com>
       From: Bob <sip:bob@biloxi.com>;tag=456248
       Call-ID: 843817637684230@998sdasdh09
       CSeq: 1826 REGISTER
       Contact: <sip:bob@192.0.2.4>
       Expires: 7200
       Content-Length: 0
```

### DIFF
--- a/include/re_sip.h
+++ b/include/re_sip.h
@@ -320,8 +320,8 @@ int  sip_contact_print(struct re_printf *pf,
 
 
 /* dialog */
-int  sip_dialog_alloc(struct sip_dialog **dlgp,
-		      const char *uri, const char *to_uri,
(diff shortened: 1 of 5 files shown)


alfredh:
it turned out that this patch was not necessary.

instead, the from name was added in another PR:

https://github.com/creytiv/re/pull/104

PR #104, opened 2018-01-05
sipreg: add from_name (Display Name)

alfredh:


amine-amine:
The demo connot run after adding fromname

PR #106, opened 2018-01-08
aes: add support for GCM Galois Counter Mode

alfredh:
This patch adds support for AES GCM mode, using the OpenSSL backend.
The usecase is TLS and SRTP.

The GCM mode is of type AEAD and provides both encryption
and authentication (with a tag). This is different from e.g. CTR-mode which
only provides encryption.

The code is functional, and the API is a suggestion.

Design:

* Provide AAD via `aes_encr` and `aes_decr`
* Separate functions to get/verify authentication tag

Comparison of CTR and GCM modes:

|                 | AES CTR  | AES GCM  |
| --------------- | -------- | -------- |
| Encryption      | Yes      | Yes      |
| Authentication  | No       | Yes      |
| IV              | 16 bytes | 12 bytes |
| Tag length      | -        | 16 bytes |

Retest:

https://github.com/creytiv/retest/tree/aes_add_gcm

- [x] OpenSSL 1.0.1t (Raspbian armv7) commit 10fdf464b8ecce06b37b916d8dd16ed9732c4e0a
- [x] OpenSSL 1.0.2n (OSX) commit 10fdf464b8ecce06b37b916d8dd16ed9732c4e0a
- [x] OpenSSL 1.1.0g (Debian x86_64) commit 10fdf464b8ecce06b37b916d8dd16ed9732c4e0a
- [x] LibreSSL 2.6.3 (OpenBSD 6.2) commit 10fdf464b8ecce06b37b916d8dd16ed9732c4e0a
- [x] Valgrind (i686) commit 10fdf464b8ecce06b37b916d8dd16ed9732c4e0a
- [x] OpenWRT/LEDE 17.01 (mips, OpenSSL 1.0.2n) commit 10fdf464b8ecce06b37b916d8dd16ed9732c4e0a

Alfred

### DIFF
--- a/src/aes/openssl/aes.c
+++ b/src/aes/openssl/aes.c
@@ -13,12 +13,58 @@
 #include <re_aes.h>
 
 
-#ifdef EVP_CIPH_CTR_MODE
-
-
 struct aes {
 	EVP_CIPHER_CTX *ctx;
+	enum aes_mode mode;
+	bool encr;
 };
+
+
+static const EVP_CIPHER *aes_cipher(enum aes_mode mode, size_t key_bits)
+{
+	if (mode == AES_MODE_CTR) {
+
+		switch (key_bits) {
+
+		case 128: return EVP_aes_128_ctr();
+		case 192: return EVP_aes_192_ctr();
+		case 256: return EVP_aes_256_ctr();
+		default:
+			return NULL;
+		}
+	}
+	else if (mode == AES_MODE_GCM) {
+
+		switch (key_bits) {
+
+		case 128: return EVP_aes_128_gcm();
+		case 256: return EVP_aes_256_gcm();
+		default:
+			return NULL;
+		}
+	}
+	else {
+		return NULL;
+	}
+}
+
+
+static inline bool set_crypt_dir(struct aes *aes, bool encr)
+{
+	if (aes->encr != encr) {
+
+		/* update the encrypt/decrypt direction */
+		if (!EVP_CipherInit_ex(aes->ctx, NULL, NULL,
+				       NULL, NULL, encr)) {
+			ERR_clear_error();
+			return false;
+		}
+
+		aes->encr = encr;
+	}
+
+	return true;
+}
 
 
 static void destructor(void *arg)
@@ -38,7 +84,7 @@ static void destructor(void *arg)
 
 int aes_alloc(struct aes **aesp, enum aes_mode mode,
 	      const uint8_t *key, size_t key_bits,
-	      const uint8_t iv[AES_BLOCK_SIZE])
+	      const uint8_t *iv)
 {
 	const EVP_CIPHER *cipher;
 	struct aes *st;
@@ -47,12 +93,16 @@ int aes_alloc(struct aes **aesp, enum aes_mode mode,
 	if (!aesp || !key)
 		return EINVAL;
 
-	if (mode != AES_MODE_CTR)
+	cipher = aes_cipher(mode, key_bits);
+	if (!cipher)
 		return ENOTSUP;
 
 	st = mem_zalloc(sizeof(*st), destructor);
 	if (!st)
 		return ENOMEM;
+
+	st->mode = mode;
+	st->encr = true;
 
 #if OPENSSL_VERSION_NUMBER >= 0x10100000L
 	st->ctx = EVP_CIPHER_CTX_new();
@@ -72,17 +122,6 @@ int aes_alloc(struct aes **aesp, enum aes_mode mode,
 	EVP_CIPHER_CTX_init(st->ctx);
 #endif
 
-	switch (key_bits) {
-
-	case 128: cipher = EVP_aes_128_ctr(); break;
-	case 192: cipher = EVP_aes_192_ctr(); break;
-	case 256: cipher = EVP_aes_256_ctr(); break;
-	default:
-		re_fprintf(stderr, "aes: unknown key: %zu bits\n", key_bits);
-		err = EINVAL;
-		goto out;
-	}
-
 	r = EVP_EncryptInit_ex(st->ctx, cipher, NULL, key, iv);
 	if (!r) {
 		ERR_clear_error();
@@ -99,7 +138,7 @@ int aes_alloc(struct aes **aesp, enum aes_mode mode,
 }
 
 
-void aes_set_iv(struct aes *aes, const uint8_t iv[AES_BLOCK_SIZE])
+void aes_set_iv(struct aes *aes, const uint8_t *iv)
 {
 	int r;
 
@@ -116,8 +155,11 @@ int aes_encr(struct aes *aes, uint8_t *out, const uint8_t *in, size_t len)
 {
 	int c_len = (int)len;
 
-	if (!aes || !out || !in)
+	if (!aes || !in)
 		return EINVAL;
+
+	if (!set_crypt_dir(aes, true))
+		return EPROTO;
 
 	if (!EVP_EncryptUpdate(aes->ctx, out, &c_len, in, (int)len)) {
 		ERR_clear_error();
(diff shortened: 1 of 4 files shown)


src/aes/openssl/aes.c
```diff
@@ -128,6 +153,147 @@ int aes_encr(struct aes *aes, uint8_t *out, const uint8_t *in, size_t len)
...
+
+	switch (aes->mode) {
+
+	case AES_MODE_GCM:
+		break;
+	default:
+		return ENOTSUP;
+	}
+
+	if (!EVP_CipherUpdate(aes->ctx, NULL, &tmplen, aad, (int)aad_len)) {
```
lgrahl: `size_t` to `int` cast should probably be checked.

src/aes/openssl/aes.c
```diff
@@ -128,6 +153,147 @@ int aes_encr(struct aes *aes, uint8_t *out, const uint8_t *in, size_t len)
...
+ * @param taglen Length of Authentication tag
+ * @param in     Plaintext input (optional)
+ * @param len    Length of plaintext input (optional)
+ *
+ * @return 0 if success, otherwise errorcode
+ */
+int aes_auth_encr(struct aes *aes, uint8_t *out, uint8_t *tag, size_t taglen,
+		  const uint8_t *in, size_t len)
+{
+	int c_len = (int)len;
```
lgrahl: Same as above.

src/aes/openssl/aes.c
```diff
@@ -128,6 +153,147 @@ int aes_encr(struct aes *aes, uint8_t *out, const uint8_t *in, size_t len)
...
+ * @param taglen Length of Authentication tag
+ *
+ * @return 0 if success, otherwise errorcode
+ *
+ * @retval EAUTH if authentication failed
+ */
+int aes_auth_decr(struct aes *aes, uint8_t *out, const uint8_t *in, size_t len,
+		  const uint8_t *tag, size_t taglen)
+{
+	int c_len = (int)len;
```
lgrahl: Same as above.

src/aes/openssl/aes.c
```diff
@@ -128,6 +153,147 @@ int aes_encr(struct aes *aes, uint8_t *out, const uint8_t *in, size_t len)
...
+
+	if (!EVP_EncryptFinal_ex(aes->ctx, NULL, &tmplen))
+		goto error;
+
+	(void)tmplen;  /* not used */
+
+	if (tag && taglen) {
+
+		if (!EVP_CIPHER_CTX_ctrl(aes->ctx, EVP_CTRL_GCM_GET_TAG,
+					 (int)taglen, tag))
```
lgrahl: Same as above.

src/aes/openssl/aes.c
```diff
@@ -128,6 +153,147 @@ int aes_encr(struct aes *aes, uint8_t *out, const uint8_t *in, size_t len)
...
+	/* update the encrypt/decrypt direction */
+	if (!EVP_CipherInit_ex(aes->ctx, NULL, NULL, NULL, NULL, false))
+		goto error;
+
+	if (!EVP_DecryptUpdate(aes->ctx, out, &c_len, in, (int)len))
+		goto error;
+
+	/* Set expected tag value. */
+	if (!EVP_CIPHER_CTX_ctrl(aes->ctx, EVP_CTRL_GCM_SET_TAG,
+				 (int)taglen, (void *)tag))
```
lgrahl: Same as above.

alfredh:
Hi @lgrahl many thanks for your input

Libsodium supports AES-GCM, and they have two modes combined
and detached (separate TAG or not):

https://download.libsodium.org/doc/secret-key_cryptography/aes-256-gcm.html

the libsodium API is completely stateless. With AEAD-ciphers such as GCM you are
supposed to verify the tag before decryption. How does this work with streaming ?
If you are streaming a very large file (100 GB) you have to decrypt and auth all blocks,
and then verify the tag _before_ looking at any plaintext data.

1. Separate AAD function? Yes, agree.
2. Separate TAG parameters? Yes, agree.
3. Streaming support ? Not sure yet.

Alfred

lgrahl:
Yeah, streamed use of these kinds of constructions isn't trivial when decrypting. Basically, one would have to discard all received (and potentially stored) data in the end in case the TAG/MAC check fails.

If there's no immediate use case for you or someone else, I guess it would be fine to ignore streaming support for now. (AES-GCM wouldn't really be my first choice for large encrypted file transfers anyway. :) )

andreygursky:
> AES-GCM wouldn't really be my first choice for large encrypted file transfers anyway.

What about doing like TLS? Just specify chunk size (variable up to ... or fixed e.g. 64K) and encrypt a piece, than increment IV, encrypt again and so on. Many of us use wget/curl for downloading (very) large files over HTTPS.

src/aes/openssl/aes.c
alfredh: do you mean that the casting is not needed ?

many openssl functions are using `int` type for counting bytes:

```
int EVP_CipherUpdate(EVP_CIPHER_CTX *ctx, unsigned char *out, int *outl,
                     const unsigned char *in, int inl);
```

alfredh:
Thank you all for comments.

with the current approach, there is some duplicated code in `aes_encr` and `aes_auth_encr`,
mainly calls to OpenSSL functions.

another option is to add functions that are specific to AEAD (GCM):

```
int  aes_add_aad(struct aes *aes, const uint8_t *aad, size_t aad_len);

int  aes_encr_get_tag(struct aes *aes, uint8_t *tag, size_t taglen);

int  aes_decr_verify_tag(struct aes *aes, const uint8_t *tag, size_t taglen);
```

src/aes/openssl/aes.c
lgrahl: The casting is needed but I'd do checks such as `aad_len > INT_MAX`.

alfredh:
Hei @richaas do you have some time to review the API suggestion?


Alfred

richaas:
I suggest the following:
- Use current `aes_encr()` and `aes_decr()` also with AEAD type modes. Multiple calls may be performed before calculating auth tag (encrypt) or verifying authenticity (decrypt).
- Use current `aes_encr()` and `aes_decr()` to set AAD, simply by allowing the `out` argument to be `NULL`. Again, multiple calls may be performed.
- These new, AEAD related API methods are needed:
```
int aes_get_authtag(struct aes *aes, uint8_t *tag, size_t taglen);
int aes_authenticate(struct aes *aes, const uint8_t *tag, size_t taglen);
```
Here is some nice documentation with example code:
https://wiki.openssl.org/index.php/EVP_Authenticated_Encryption_and_Decryption

richaas:
A separate note regarding length of IV. When adding GCM, I suggest we change:
`const uint8_t iv[AES_BLOCK_SIZE]`
to
`const uint8_t *iv`

GCM also support configurable IV length, but we may stick to the default if we would like to keep the current API backwards compatible. Otherwise we would need to include `size iv_len` parameter to `aes_alloc()` and `aes_set_iv()`.

alfredh:
Hi @richaas many thanks for your comments.

I have now updated the patch with your suggested API. The code became
smaller and more clear.

Could you please review the complete patch again?

Testing was done with `retest` and the results are in the first comment (at the top).

Alfred

richaas:
Some more comments:

- I think we should create a new function that maps mode/bits to cipher. Check if a mapping table or nested switch is the best approach. This should be called before creating state, so we can return directly if this fails. The debug printfs should be removed.

`static const EVP_CIPHER *aes_cipher(enum aes_mode mode, size_t key_bits)`

- It seems suboptimal to set direction for each `aes_decr()` call. I suggest we set this once during alloc instead, by adding a new argument.

- We need dummy tag functions if EVP_CIPH_CTR_MODE is not defined. Or maybe we should stop supporting this case?

- I suggest we use switch in the tag functions, in order to be better prepared for future modes:

```
int aes_get_authtag(struct aes *aes, uint8_t *tag, size_t taglen)
{
	int tmplen;

	if (!aes || !tag || !taglen)
		return EINVAL;

	switch (aes->mode) {

	case AES_MODE_GCM:
		if (!EVP_EncryptFinal_ex(aes->ctx, NULL, &tmplen)) {
			ERR_clear_error();
			return EPROTO;
		}

		if (!EVP_CIPHER_CTX_ctrl(aes->ctx, EVP_CTRL_GCM_GET_TAG,
					 (int)taglen, tag)) {
			ERR_clear_error();
			return EPROTO;
		}

		return 0;

	default:
		return ENOTSUP;
	}
}


int aes_authenticate(struct aes *aes, const uint8_t *tag, size_t taglen)
{
	int tmplen;

	if (!aes || !tag || !taglen)
		return EINVAL;

	switch (aes->mode) {

	case AES_MODE_GCM:
		if (!EVP_CIPHER_CTX_ctrl(aes->ctx, EVP_CTRL_GCM_SET_TAG,
					 (int)taglen, (void *)tag)) {
			ERR_clear_error();
			return EPROTO;
		}

		if (EVP_DecryptFinal_ex(aes->ctx, NULL, &tmplen) <= 0) {
			ERR_clear_error();
			return EAUTH;
		}

		return 0;

	default:
		return ENOTSUP;
	}
}
```

alfredh:
The Openssl backend for AES supports 2 different APIs:

1. EVP api, this is the recommended API to use. AES was supported since 1.0.1

2. AES api, this is a very old API going back to 0.9.8

We dont support openssl 0.9.8 any more, libre now only supports 1.0.1 or later.

from openssl.org:

```
Note: The latest stable version is the 1.1.0 series. The 1.0.2 series is our Long Term Support (LTS) release, supported until 31st December 2019. The 0.9.8, 1.0.0 and 1.0.1 versions are now out of support and should not be used.
```

from my perspective, we can remove usage of old "AES" API, and only use the EVP api.

richaas:
Agree, let's remove the old "AES" API.

alfredh:
all comments should be fixed now, except the direction issue.

when the AES context is created, it does not know about encrypt/decrypt direction.
the same context can be used for encryption or decryption, provided that the IV is
reset. This is working with CTR, but it is not working with GCM.

it is possible to add a flag `bool encr` to `aes_alloc` that sets the desired direction
at alloc time. This change also means that we have to change the SRTP code that
is using the same AES context for encrypt/decrypt.

The SRTP code can allocate the AES context for each encryption/decrypt round,
since the IV is always reset.

alfredh:
one solution could be to add a function, for changing the encrypt/decrypt direction:

for example:

```
void aes_set_direction(struct aes *aes, bool encr)
{
	/* update the encrypt/decrypt direction */
	if (!EVP_CipherInit_ex(aes->ctx, NULL, NULL, NULL, NULL, encr)) {
		re_printf("set direction error\n");
		ERR_clear_error();
	}
}
```

richaas:
I had a similar idea as your last commit, using and inline function to set direction; patch below:

[aes.diff.txt](https://github.com/creytiv/re/files/1678038/aes.diff.txt)

BTW: is it ok to call `EVP_EncryptInit_ex()` for both directions??

alfredh:
the context must be initialized to something, either encrypt (1) or decrypt (0):

```
	r = EVP_EncryptInit_ex(st->ctx, cipher, NULL, key, iv);
	if (!r) {
		ERR_clear_error();
		err = EPROTO;
	}

or

int EVP_CipherInit(EVP_CIPHER_CTX *ctx, const EVP_CIPHER *cipher,
                   const unsigned char *key, const unsigned char *iv,
                   int enc);
```

where "int enc" is either 1 or 0.

setting the crypt direction automatically I think is a flexible solution,
it does not need any changes to application code (such as SRTP).

richaas:
In `aes_set_iv()` we should not change direction to encrypt, but rather keep the current direction, so we need to replace current `EVP_EncryptInit_ex()` call with:

`EVP_CipherInit_ex(aes->ctx, NULL, NULL, NULL, iv, -1);`

PR #109, opened 2018-01-18
Allow API users to override the target of an in-dialog message

rkd-msw:
Hi,

In our application, we have reasonably sophisticated DNS requirements, so we're trying to avoid using re's built-in DNS, doing DNS resolving ourselves, and just passing re the IP addresses/ports to use.

In order to do this for in-dialog requests, we found we needed to add some methods:

- exposing the Route header on a dialog (learned from Record-Route) so we can resolve it
- a `sip_drequestf_targeted` alternative to `sip_drequestf`, which lets us specify a target rather than relying on the dialog route

I've also added a warning when we try to do DNS resolution with no DNS client, as this is tricky to diagnose otherwise.

### DIFF
--- a/include/re_sip.h
+++ b/include/re_sip.h
@@ -281,6 +281,11 @@ int sip_drequestf(struct sip_request **reqp, struct sip *sip, bool stateful,
 		  const char *met, struct sip_dialog *dlg, uint32_t cseq,
 		  struct sip_auth *auth, sip_send_h *sendh, sip_resp_h *resph,
 		  void *arg, const char *fmt, ...);
+int sip_drequestf_targeted(struct sip_request **reqp, struct sip *sip,
+		  bool stateful, const char *met, struct sip_dialog *dlg,
+		  const struct uri *route, uint32_t cseq,
+		  struct sip_auth *auth, sip_send_h *sendh, sip_resp_h *resph,
+		  void *arg, const char *fmt, ...);
 void sip_request_cancel(struct sip_request *req);
 bool sip_request_loops(struct sip_loopstate *ls, uint16_t scode);
 void sip_loopstate_reset(struct sip_loopstate *ls);
@@ -336,6 +341,7 @@ bool sip_dialog_established(const struct sip_dialog *dlg);
 bool sip_dialog_cmp(const struct sip_dialog *dlg, const struct sip_msg *msg);
 bool sip_dialog_cmp_half(const struct sip_dialog *dlg,
 			 const struct sip_msg *msg);
+const struct uri *sip_dialog_route(const struct sip_dialog *dlg);
 
 
 /* msg */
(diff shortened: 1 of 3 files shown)


amine-amine:
Good afternoon, iam a student and i want to implement a sip client in fpgaboard, so i need help please, the demo works well with re_0.5.6 and not re-0.5.7 this my first problem, the second problem is when i work on the fpga board with the demo and re-0.5.6 it complain (invalid argument) afollowit and i find that it is (*(srvc)) in the function query, anorher problem is the struct dnsc is done in client.c so when i try to do a new pragram using ur beatiful library the struct the program coyldnt know tje struct dnsc because is was definedin client.c

amine-amine:
Thk sooooo much for helping me

src/dns/client.c
```diff
@@ -749,8 +749,10 @@ int dnsc_query(struct dns_query **qp, struct dnsc *dnsc, const char *name,
 	       uint16_t type, uint16_t dnsclass,
 	       bool rd, dns_query_h *qh, void *arg)
 {
-	if (!dnsc)
+	if (!dnsc) {
+		DEBUG_WARNING("DNS query attempted but DNS client is unconfigured\n");
 		return EINVAL;
+	}
```
alfredh: most of the code in re does not print a warning if a required parameter is NULL.

I think it would be better to move this check/warning to the SIP stack and check for `dnsc`
explicitly there.

src/sip/request.c
```diff
@@ -847,6 +828,94 @@ int sip_drequestf(struct sip_request **reqp, struct sip *sip, bool stateful,
...
+	va_end(ap);
+
+	if (err)
+		goto out;
+
+	err = sip_drequestf_common(reqp, sip, stateful, met, dlg, route, cseq,
+				   auth, sendh, resph, arg, fmt_str);
+out:
+	if (fmt_str)
+		mem_deref(fmt_str);
```
alfredh: checking `fmt_str` is not needed, this is done by `mem_deref`

include/re_sip.h
```diff
@@ -336,6 +341,7 @@ bool sip_dialog_established(const struct sip_dialog *dlg);
 bool sip_dialog_cmp(const struct sip_dialog *dlg, const struct sip_msg *msg);
 bool sip_dialog_cmp_half(const struct sip_dialog *dlg,
 			 const struct sip_msg *msg);
+const struct uri *sip_dialog_route(const struct sip_dialog *dlg);
```
alfredh: if you want to export this function, the prototype should be removed from `src/sip/sip.h`

alfredh:
hi @rkd-msw thanks for your patch.

Could you please tell us a bit more about your DNS requirements ?
It would be interesting to know why the DNS client in libre cannot be used ..

there is one ccheck warning that must be fixed:

```
./src/dns/client.c:753: line is too wide (86 - max 79)
```

I made some minor comments to your PR.


Alfred

alfredh:
hi @rkd-msw 

did you get my message about the review comments ?

Alfred

PR #111, opened 2018-02-08
Srtp add gcm

alfredh:
This PR adds support for AES-GCM cipher in the SRTP Stack, as defined
in https://tools.ietf.org/html/rfc7714

Retest: https://github.com/creytiv/retest/tree/srtp_add_gcm

Srtp performance test: https://github.com/alfredh/srtperf/tree/srtp_add_gcm

TODO list:

- [x] Interop testing with Chrome 63 (success) via DTLS-SRTP 128/256-bits
- [x] Interop testing with Chrome 64 (success) via DTLS-SRTP 128/256-bits
- [x] Interop testing with pjsip 2.7.2 (success) via SDES 128-bit and 256-bit
- [x] test with Valgrind (success)
- [x] test on Big Endian CPU (OpenWRT MIPS) (success)

### DIFF
--- a/src/srtp/srtcp.c
+++ b/src/srtp/srtcp.c
@@ -66,8 +66,7 @@ int srtcp_encrypt(struct srtp *srtp, struct mbuf *mb)
 
 		ep = 1;
 	}
-
-	if (rtcp->aes && rtcp->mode == AES_MODE_GCM) {
+	else if (rtcp->aes && rtcp->mode == AES_MODE_GCM) {
 
 		union vect128 iv;
 		uint8_t *p = mbuf_buf(mb);
@@ -212,8 +211,7 @@ int srtcp_decrypt(struct srtp *srtp, struct mbuf *mb)
 		if (err)
 			return err;
 	}
-
-	if (rtcp->aes && ep && rtcp->mode == AES_MODE_GCM) {
+	else if (rtcp->aes && ep && rtcp->mode == AES_MODE_GCM) {
 		union vect128 iv;
 		uint8_t *p;
 		size_t hdr_len;
--- a/src/srtp/srtcp.c
+++ b/src/srtp/srtcp.c
@@ -216,22 +216,26 @@ int srtcp_decrypt(struct srtp *srtp, struct mbuf *mb)
 	if (rtcp->aes && ep && rtcp->mode == AES_MODE_GCM) {
 		union vect128 iv;
 		uint8_t *p;
+		size_t hdr_len;
 		size_t tag_start;
 		size_t pld_len;
 
+		hdr_len   = pld_start - start;
 		tag_start = mb->end - GCM_TAGLEN;
 		pld_len   = tag_start - pld_start;
 
 		mb->pos = pld_start;
 		p = mbuf_buf(mb);
 
+		if (mbuf_get_left(mb) < GCM_TAGLEN)
+			return EBADMSG;
+
 		srtp_iv_calc_gcm(&iv, &rtcp->k_s, ssrc, ix);
 
 		aes_set_iv(rtcp->aes, iv.u8);
 
 		/* The RTP Header is Associated Data */
-		err = aes_decr(rtcp->aes, NULL, &mb->buf[start],
-				  pld_start - start);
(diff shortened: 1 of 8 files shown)


alfredh:
hi @richaas 

do you have some time to review this patch ?

Alfred

src/srtp/srtcp.c
```diff
@@ -178,6 +213,47 @@ int srtcp_decrypt(struct srtp *srtp, struct mbuf *mb)
 			return err;
 	}
 
+	if (rtcp->aes && ep && rtcp->mode == AES_MODE_GCM) {
```
richaas: else if ?

src/srtp/srtp.c
```diff
@@ -315,6 +379,50 @@ int srtp_decrypt(struct srtp *srtp, struct mbuf *mb)
 		if (err)
 			return err;
 	}
+	else if (comp->aes && comp->mode == AES_MODE_GCM) {
+
```
richaas: I suggest we replace the entire if block with this code for clarity, simplicity and consistence with RTCP code:
```
		union vect128 iv;
		uint8_t *p = mbuf_buf(mb);
		size_t tag_start;

		srtp_iv_calc_gcm(&iv, &comp->k_s, strm->ssrc, ix);

		aes_set_iv(comp->aes, iv.u8);

		/* The RTP Header is Associated Data */
		err = aes_decr(comp->aes, NULL, &mb->buf[start],
			       mb->pos - start);
		if (err)
			return err;

		if (mbuf_get_left(mb) < GCM_TAGLEN)
			return EBADMSG;

		tag_start = mb->end - GCM_TAGLEN;

		err = aes_decr(comp->aes, p, p, tag_start - mb->pos);
		if (err)
			return err;

		err = aes_authenticate(comp->aes, &mb->buf[tag_start],
				       GCM_TAGLEN);
		if (err)
			return err;

		mb->end = tag_start;

		/*
		 * 3.3.2.  Replay Protection
		 *
		 * Secure replay protection is only possible when
		 * integrity protection is present.
		 */
		if (!srtp_replay_check(&strm->replay_rtp, ix))
			return EALREADY;

```

src/srtp/srtp.h
```diff
@@ -40,8 +40,11 @@ struct srtp_stream {
 struct srtp {
 	struct comp {
 		struct aes *aes;    /**< AES Context                       */
+		enum aes_mode mode; /**< AES encryption mode               */
 		struct hmac *hmac;  /**< HMAC Context                      */
 		union vect128 k_s;  /**< Derived salting key (14 bytes)    */
+
+		// NOTE: only for HMAC-SHA1:
```
richaas: remove note?

richaas:
Ready to be merged now, or want to do more testing first?

alfredh:
@richaas thanks for the review.

The code is now a good starting point, but needs more testing I think.

alfredh:
I did some more interop testing, looks good so far. The top-level post is updated.

I am wondering if we can simplify this piece of code:

```
	if (comp->aes && comp->mode == AES_MODE_CTR) {
		union vect128 iv;
		uint8_t *p = mbuf_buf(mb);

		srtp_iv_calc(&iv, &comp->k_s, strm->ssrc, ix);

		aes_set_iv(comp->aes, iv.u8);
		err = aes_encr(comp->aes, p, p, mbuf_get_left(mb));
		if (err)
			return err;
	}
	else if (comp->aes && comp->mode == AES_MODE_GCM) {
		union vect128 iv;
		uint8_t *p = mbuf_buf(mb);
		uint8_t tag[GCM_TAGLEN];

		srtp_iv_calc_gcm(&iv, &comp->k_s, strm->ssrc, ix);

		aes_set_iv(comp->aes, iv.u8);
```

there is some duplicated code there, I am wondering if it would be better
to just have one if (comp->aes) block and within that block check for CTR or GCM ..

richaas:
I would not like one extra level of indentation there.

alfredh:
hi @richaas 

This PR should be ready to merge to master now. Please merge
if you want.

The SRTPERF program has also been updated to support GCM.
It compares all packets encrypted between libsrtp (reference)
and libre, with success.

https://github.com/alfredh/srtperf/tree/srtp_add_gcm

Alfred

PR #121, opened 2018-03-20
Changed mechanics of choosing network adapter for Windows

Encamy:
Hello
If there are two or more network adapters installed in the system then re will connect to first adapter.
This adapter can be, for example, a virtual adapter or etc. Sometimes this leads to the fact that there is no connection between the re and the server.

I've renamed _net_hostaddr()_ to _net_hostaddr_win32()_ because it affects only Windows.

Previously _net_hostaddr()_ was using _gethostbyname()_. Now it is deprecated and it's better to use _GetAddrInfo()_, _GetAddrInfoA()_ or _GetAddrInfoB()_

Now _net_hostaddr_win32()_ is getting ip and address family of all adapters installed in the system. Then it checks which address family is matching with _af_.
`static int net_hostaddr_win32(int af, struct sa *ip)`
If match was found then it executing _sa_set_in()_

### DIFF
--- a/src/net/net.c
+++ b/src/net/net.c
@@ -24,6 +24,7 @@
 #include <re_dbg.h>
 
 
+#if defined(WIN32)
 /**
  * Get the IP address of the host
  *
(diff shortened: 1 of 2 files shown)


alfredh:
hi, thanks for your PR. here are some comments:

* platform specific code should be located under a subfolder, such as:
  `src/net/win32/...`

* compiling with i686-w64-mingw32 (gcc 7.3.) gives a warning:

```
src/net/net.c: In function ‘net_hostaddr_win32’:
src/net/net.c:60:37: warning: initialization from incompatible pointer type [-Wincompatible-pointer-types]
    struct sockaddr_in * addr_ipv4 = ptr->ai_addr;
```

* your code is not working with IPv6. in my test when asking for AF_INET6
  it returns the address 0.0.0.0

* some more inline comments ..


Alfred

src/net/net.c
```diff
@@ -32,28 +33,38 @@
...
 
 	if (-1 == gethostname(hostname, sizeof(hostname)))
 		return errno;
 
-	he = gethostbyname(hostname);
-	if (!he)
-		return ENOENT;
+	struct addrinfo *result = NULL;
+	struct addrinfo *ptr = NULL;
+	bool ip_family_match = false;
```
alfredh: variables should be declared at the beginning of the function.

src/net/net.c
```diff
@@ -32,28 +33,38 @@
...
 
-	if (af != he->h_addrtype)
-		return EAFNOSUPPORT;
+	/*gethostbyname is deprecated. It's better to use GetAddrInfo instead*/
+	GetAddrInfo(hostname, NULL, NULL, &result);
```
alfredh: the return value of GetAddrInfo must be checked.

src/net/net.c
```diff
@@ -32,28 +33,38 @@
...
-	memcpy(&in, he->h_addr_list[0], sizeof(in));
-	sa_set_in(ip, ntohl(in.s_addr), 0);
+	/*Get entry which have similar to af ip family (for example IPv4)*/
+	/*Previously net_hostaddr set ip address of the first adapter*/
+	for (ptr = result; ptr != NULL; ptr = ptr->ai_next) {
+		if (ptr->ai_family == af)
+		{
+			ip_family_match = true;
+			struct sockaddr_in * addr_ipv4 = ptr->ai_addr;
+			sa_set_in(ip, ntohl(addr_ipv4->sin_addr.s_addr), 0);
```
alfredh: this will only work for IPv4. it needs to support both IPv4 and IPv6.

you can use e.g. `sa_set_in6` or `sa_set_sa` functions.

alfredh:
also, with this PR it breaks one test in baresip selftest when compiled with
mingw and running under wine:

https://github.com/alfredh/baresip-win32

alfredh:
hi @Encamy did you have a chance to look at my comments ?

Encamy:
Hello @alfredh 
I apologize for not responding. Now I'm busy with another task. As soon as I can, I will correct mistakes. Should I close the PR?

alfredh:
please close the PR and create a new one when you have time to work on the issue.

Alfred

Encamy:
Hello @alfredh,
Finally I've made changes according to your comments.
Should I create a new pull request?

alfredh:
yes, that would be nice :)

PR #122, opened 2018-03-23
mqueue: set non-blocking mode for read/write file descriptors

richaas:
In order to avoid pipe "constipation", we need to set the read/write file descriptors in non-blocking mode. Otherwise write will block if pipe is full, and read will block if pipe is empty.

### DIFF
--- a/src/mqueue/mqueue.c
+++ b/src/mqueue/mqueue.c
@@ -10,6 +10,7 @@
 #include <re_fmt.h>
 #include <re_mem.h>
 #include <re_main.h>
+#include <re_net.h>
 #include <re_mqueue.h>
 #include "mqueue.h"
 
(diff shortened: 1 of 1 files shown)


alfredh:
the patch itself looks fine, but here are some questions:

1. Write; this changes the API semantics. The original code will block until the pipe has
  free space, meaning that the delivery of the event from thread X to re_main is
  guaranteed. With this patch applied, the function `mqueue_push` may return `EAGAIN`
  if the pipe is full, and it is not sure what the application should do in this case .. resend it ?

2. Read; since the event handler is triggered by an async event from re_main (poll, epoll, ...)
  does it matter if the FD is blocking or non-blocking ? I mean, is it possible that we receive
  a FD_READ event, and there is nothing to read ?


apart from that, the patch is working. briefly tested on OpenBSD 6.2 and Arch Linux.

Alfred

richaas:
libre is build around the principle of non-blocking, asynchronous operations. That should apply to all function calls, including mqueue. No function call should ever block.

alfredh:
great :)

I have merged the patch to master.

PR #127, opened 2018-04-05
json/utf8: fix unescaping of unicode code points

richaas:
Unicode code points above U+007F are currently not unescaped correctly. This patch fixes that, so that all code points (U+0000 - U+10FFFF) are properly unescaped in UTF-8 format.

@alfredh please review, test and merge.

### DIFF
--- a/src/fmt/unicode.c
+++ b/src/fmt/unicode.c
@@ -86,6 +86,7 @@ int utf8_encode(struct re_printf *pf, const char *str)
  */
 int utf8_decode(struct re_printf *pf, const struct pl *pl)
 {
+	int uhi = -1;
 	size_t i;
 
 	if (!pf)
@@ -101,7 +102,9 @@ int utf8_decode(struct re_printf *pf, const struct pl *pl)
 
 		if (ch == '\\') {
 
-			uint16_t u = 0;
+			unsigned u = 0;
+			char ubuf[4];
+			size_t ulen;
 
 			++i;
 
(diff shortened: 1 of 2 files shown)


alfredh:
do you have some example UTF8/UTF16 strings that we can use for testcases ?

richaas:
Å - \u00c5
€ - \u20ac
𝄞 - \ud834\udd1e
𒌷  𒉌  𒉡  𒀀 - \ud808\udf37  \ud808\ude4c  \ud808\ude61  \ud808\udc00

alfredh:
I spent some time to refresh my Unicode knowledge. Unicode is fun!
The patch looks good.

I added some tests here:

https://github.com/creytiv/retest/tree/json_utf8

it has been run-time tested on OSX and OpenWRT (MIPS/Big Endian) with success.

src/fmt/unicode.c
```diff
@@ -147,22 +150,81 @@ int utf8_decode(struct re_printf *pf, const struct pl *pl)
...
-					if (err)
-						return err;
+				/* UTF-16 surrogate pair */
+				if (u >= 0xd800 && u <= 0xdbff) {
+					uhi = (u - 0xd800) * 0x400;
+					continue;
 				}
+				else if (u >= 0xdc00 && u <= 0xdfff) {
+					if (uhi < 0)
+						continue;
```
alfredh: this code seems to decode a UTF-16 surrogate pair (first high, then low).

if a corrupt Unicode string is detected, should we just ignore it or stop the decoding
with an error ?

for example this string would be corrupt:

```\\uD834\\uD834```

richaas: IMO it is best to just ignore invalid code points, so that all the part of the string that is valid gets decoded properly. I think that is the best and most robust behaviour. This patch is written with that in mind.

alfredh:
great, I merged the patch to master.

will also merge the retest patch.

Alfred

Issue #129, opened 2018-04-15
No rules to build library with 64-bit Mingw


AlexanderplUs:
When I tried to build libre with 64-bit Mingw script `mk/re.mk` didn't detect that it is a WIN32 platform and build failed. As I checked mk/re.mk I found that it check output of `gcc -dumpmachine`. For 64-bit Mingw it gives `x86_64-w64-mingw32`

So, after I added this value to check in `mk/re.mk` libre compiles successfuly. Please consider to add support for 64-bit Mingw to the build scripts. Here is diff with my fix:
```
--- re.mk.orign 2018-04-15 15:56:52.531056000 +0400
+++ re.mk       2018-04-15 15:57:02.312273200 +0400
@@ -205,7 +205,7 @@


 ifneq ($(strip $(filter i386-mingw32 i486-mingw32 i586-mingw32msvc \
-       i686-w64-mingw32 mingw32, \
+       i686-w64-mingw32 x86_64-w64-mingw32 mingw32, \
        $(MACHINE))),)
        OS   := win32
 ifeq ($(MACHINE), mingw32)
```

alfredh:
you can also have a look at this project for how to compile re/rem for mingw:

https://github.com/alfredh/baresip-win32

your change looks fine, would you like to create a PR for it ?

richaas:
PR #131 merged to master

PR #131, opened 2018-04-17
Added support for 64-bit MINGW

AlexanderplUs:
Fixed an issue https://github.com/creytiv/re/issues/129


alfredh:
Hi @AlexanderplUs thanks for your patch.

@richaas the patch looks fine to me. okay to merge ?

PR #138, opened 2018-05-14
Changed mechanics of choosing network adapter for Windows

Encamy:
Hello
Previous [pull request](https://github.com/creytiv/re/pull/121) contained some mistakes. I hope they are all fixed now :)

> If there are two or more network adapters installed in the system then re will connect to first adapter.
This adapter can be, for example, a virtual adapter or etc. Sometimes this leads to the fact that there is no connection between the re and the server.

@alfredh would you like to check this code please?


mk/win32/re.vcxproj
```diff
@@ -158,6 +158,7 @@
     <ClCompile Include="..\..\src\net\rt.c" />
     <ClCompile Include="..\..\src\net\sock.c" />
     <ClCompile Include="..\..\src\net\sockopt.c" />
+    <ClCompile Include="..\..\src\net\win32\hostaddr.c" />
```
alfredh: please also update src/net/mod.mk so that the code can be compiled with mingw32.

src/net/win32/hostaddr.c
```diff
@@ -0,0 +1,54 @@
...
+		return errno;
+
+	err = GetAddrInfo(hostname, NULL, NULL, &result);
+	if (err)
+		return err;
+	if (!result)
+		return ENOENT;
+
+	for (ptr = result; ptr != NULL; ptr = ptr->ai_next)
+	{
```
alfredh: please move the curly braces to the previous line

src/net/win32/hostaddr.c
```diff
@@ -0,0 +1,54 @@
...
+		if (ptr->ai_family == af)
+		{
```
alfredh: please move the curly braces to the previous line

src/net/win32/hostaddr.c
```diff
@@ -0,0 +1,54 @@
...
+			ip_family_match = true;
```
alfredh: should this also break for the for-loop, so that the first address is used ?

what if the host has many ip-addresses, which one should be used ?

src/net/win32/hostaddr.c
```diff
@@ -0,0 +1,54 @@
...
+			struct sockaddr * addr = ptr->ai_addr;
+			sa_set_sa(ip, addr);
```
alfredh: you must check the return value of `sa_set_sa`.

in case of errors, either return from the function, or try the next entry in the list.

src/net/win32/hostaddr.c
```diff
@@ -0,0 +1,54 @@
...
+		}
+	}
+
+	if (!ip_family_match)
+		return EAFNOSUPPORT;
+	return 0;
+}
```
alfredh: please add a newline here.

src/net/win32/hostaddr.c
```diff
@@ -0,0 +1,54 @@
...
+	char hostname[256];
+	struct addrinfo *result = NULL;
+	struct addrinfo *ptr = NULL;
+	bool ip_family_match = false;
+	int err = 0;
+
+	if (-1 == gethostname(hostname, sizeof(hostname)))
+		return errno;
+
+	err = GetAddrInfo(hostname, NULL, NULL, &result);
```
alfredh: you can also use the 3rd parameter as "hints" to provide the AF that you are interested in.

alfredh:
hi, thanks for your patch. I added some comments.

you can also use this project to verify that it is working with Mingw32:

https://github.com/alfredh/baresip-win32

Encamy:
Hello @alfredh 
Thank you for your comments!
Here is log from [this project](https://github.com/alfredh/baresip-win32):

> retest: libre version 0.5.7 (i686/win32)
> using async polling method 'select'
> regular tests:       json: value of unknown type: <t>
> json: value of unknown type: <f>
> json: value of unknown type: <n>
> json: value of unknown type: <frue>
> json: value of unknown type: <talse>
> OK
> 
> running baresip selftest version 0.5.9 with 33 tests
> [ RUN      ] test_account
> [       OK ]
> [ RUN      ] test_aulevel
> [       OK ]
> [ RUN      ] test_call_af_mismatch
> [       OK ]
> [ RUN      ] test_call_answer
> [       OK ]
> [ RUN      ] test_call_answer_hangup_a
> [       OK ]
> [ RUN      ] test_call_answer_hangup_b
> [       OK ]
> [ RUN      ] test_call_reject
> [       OK ]
> [ RUN      ] test_call_rtp_timeout
> [       OK ]
> [ RUN      ] test_call_mediaenc
> [       OK ]
> [ RUN      ] test_call_multiple
> [       OK ]
> [ RUN      ] test_call_max
> [       OK ]
> [ RUN      ] test_call_dtmf
> [       OK ]
> [ RUN      ] test_call_aulevel
> [       OK ]
> [ RUN      ] test_call_progress
> [       OK ]
> [ RUN      ] test_call_format_float
> [       OK ]
> [ RUN      ] test_call_video
> [       OK ]
> [ RUN      ] test_video
> [       OK ]
> [ RUN      ] test_cmd
> [       OK ]
> [ RUN      ] test_cmd_long
> [       OK ]
> [ RUN      ] test_contact
> [       OK ]
> [ RUN      ] test_cplusplus
> [       OK ]
> [ RUN      ] test_event
> [       OK ]
> [ RUN      ] test_message
> [       OK ]
> [ RUN      ] test_mos
> [       OK ]
> [ RUN      ] test_network
> [       OK ]
> [ RUN      ] test_play
> [       OK ]
> [ RUN      ] test_ua_alloc
> [       OK ]
> [ RUN      ] test_ua_options
> [       OK ]
> [ RUN      ] test_ua_register
> [       OK ]
> [ RUN      ] test_ua_register_dns
> [       OK ]
> [ RUN      ] test_ua_register_auth
> [       OK ]
> [ RUN      ] test_ua_register_auth_dns
> [       OK ]
> [ RUN      ] test_uag_find_param
> [       OK ]
> OK. 33 tests passed successfully
> 

You've commented: 
>what if the host has many ip-addresses, which one should be used ?

It seems to me that the user can choose this option
But for simplicity, first ip-address can be used
Another option is get the gateway for every adapter and check for != 0.0.0.0 (for example virtual adapter have this gateway by default)

About "hints" parameter:
I don't think that it is necessary

Also I can combine all my commits into one (to prevent littering of your history)

src/net/net.c
```diff
@@ -23,39 +23,6 @@
...
-		return EAFNOSUPPORT;
-
-	/* Get the first entry */
-	memcpy(&in, he->h_addr_list[0], sizeof(in));
-	sa_set_in(ip, ntohl(in.s_addr), 0);
-
-	return 0;
-}
-
-
```
alfredh: this is a publicly exported function, and the prototype is defined in `re_net.h`:

```
int  net_hostaddr(int af, struct sa *ip);
```

after this patch, it will only be available for win32. all public functions should be available
for all platforms.

if we want to completely remove it from the API, we must also remove the prototype.

the function `gethostbyname` only works with IPv4 and is deprecated on some platforms.
libre should support both IPv4 and IPv6 properly so I guess it is a good thing to remove it...

src/net/win32/hostaddr.c
```diff
@@ -0,0 +1,56 @@
...
+	if (err)
+		return err;
+	if (!result)
+		return ENOENT;
+
+	for (ptr = result; ptr != NULL; ptr = ptr->ai_next){
+		if (ptr->ai_family == af){
```
alfredh: please add a space before the curly brackets, like this:

`af) {`

src/net/win32/hostaddr.c
```diff
@@ -0,0 +1,56 @@
...
+			struct sockaddr * addr = ptr->ai_addr;
```
alfredh: the pointer star should be aligned towards the variable name:

`struct sockaddr *addr`

src/net/win32/hostaddr.c
```diff
@@ -0,0 +1,56 @@
...
+				break;
+			}
+		}
+	}
+
+	if (!ip_family_match)
+		return EAFNOSUPPORT;
+
+	return 0;
+}
```
alfredh: please add a newline here.

alfredh:
I tested your patch on a Linux box with mingw32/wine. The IPv6 address
is provided via miredo/teredo.

the output is here:

```
Interfaces:
{00000001-0000-0000-0000-000000000000}:  127.0.0.1
{00000001-0000-0000-0000-000000000000}:  ::1
{00000002-0000-0000-0000-000000000000}:  10.0.0.2
{00000002-0000-0000-0000-000000000000}:  fe80::d63d:7eff:fef1:3048
{00000003-0000-0000-0000-000000000000}:  2001:0:53aa:64c:1cd8:7097:ab2f:8e7e
{00000003-0000-0000-0000-000000000000}:  fe80::ffff:ffff:ffff
{00000003-0000-0000-0000-000000000000}:  fe80::81de:cd3b:e2cc:e9cc
Default source addresses:
AF_INET:    127.0.1.1
AF_INET6:    Not available
```

so the AF_INET is picking up the wrong address, and the AF_INET6 is not available at all.

The test program is here:

```c
   {
                static const int afv[2] = {AF_INET, AF_INET6};
                struct sa addr;
                size_t i;

                re_printf("Interfaces:\n");
                net_if_apply(net_ifaddr_handler, 0);

                re_printf("Default source addresses:\n");
                for (i=0; i<ARRAY_SIZE(afv); i++) {
                        const int af = afv[i];

                        if (0 == net_default_source_addr_get(af, &addr))
                                re_printf("%s:    %j\n", net_af2name(af), &addr);
                        else
                                re_printf("%s:    Not available\n", net_af2name(af));
                }
        }
```

Do you have the possibility to test if IPv6 is working on your Windows machine?
(and also which version of Windows are you using) ?

src/net/net.c
Encamy: There is no function `gethostbyname` (may be you mixed up with `gethostname`?), in this PR it was replaced by `GetAddrInfo`
Why platform specific code should be placed in a subfolder? Should I create a header file and .c file for one function? In [net.c](https://github.com/creytiv/re/blob/master/src/net/net.c) it placed in same file, but it used only for Win32

alfredh: it is nice if platform specific code is placed in subfolders, it makes the code
more organized and easy to maintain.

these folders are already under `src/net`:

```
bsd
linux
posix
win32
```

alfredh:
hi @Encamy no worries,

please keep the PRs coming.

perhaps it is better to use the `net_if_apply` function to enumerate all the
interfaces, and then pick the best one.

PR #139, opened 2018-05-17
Add answer selected algorithm in http_digest auth response(RFC 7616)

sergius-dart:


alfredh:
hi @sergius-dart many thanks for your patch

we only support MD5 for now, and if the `algorithm` parameter is missing the default
algorithm is MD5. Do you have any SIP trace or SIP servers where the current code is
not working?

RFC 7616 adds support for SHA-256 algorithm, perhaps we should consider adding this.

Alfred

sergius-dart:
Yes - I found a server that requires this parameter.

alfredh:
I think it would be nice to see a SIP trace.

there is also a ccheck warning that must be fixed:

./src/sip/auth.c:276: line is too wide (84 - max 79)

sergius-dart:
But nonce and cnonce - random every request - and if i show tcpdump after and before patch - you don't know, mb i write bad password for "before patch". 
I sended you login+pass+server for test. You received my 2 email?

alfredh:
hi @sergius-dart 

thanks for the test account. I have tested your patch and it seems to work.

However, I think the libre code is correct and that your SIP Server is buggy.
From RFC 2617:

```
 algorithm
     A string indicating a pair of algorithms used to produce the digest
     and a checksum. If this is not present it is assumed to be "MD5".
     If the algorithm is not understood, the challenge should be ignored
     (and a different one used, if there is more than one).
```

In the libre code the `algorithm` string is missing, hence your SIP server
should handle this and treat it as "MD5".

Please report the bug to your SIP server provider.

so my conclusion now is that I am not sure if we should apply this patch or not.

Asking @richaas for a second opinion :)


Alfred

richaas:
I would prefer fixing the server that do not adhere to the spec, if that is possible.

joshhodgson:
Can I just add, I have just come across a server with the same issue

PR #142, opened 2018-05-24
Add support for Android (8.0) Oreo

### LINKED ISSUE #137: Android Oreo (8.0) support
Due to [changes](https://developer.android.com/about/versions/oreo/android-8.0-changes.html#o-pri) affecting privacy in android 8 trying to access the net.dns properties will result in the following:

```
E/libc: Access denied finding property "net.dns1"
E/libc: Access denied finding property "net.dns2"
E/libc: Access denied finding property "net.dns3"
E/libc: Access denied finding property "net.dns4"
```
Therefore for android versions >= 8 the following function will no longer work.
https://github.com/creytiv/re/blob/a3a44da128b0cd95c5edf7790bcf2e188b59d17d/src/dns/ns.c#L83

Hmart:
Fixes #137

### DIFF
--- a/src/dns/ns.c
+++ b/src/dns/ns.c
@@ -89,7 +89,7 @@ static int get_android_dns(struct sa *nsv, uint32_t *n)
 	int err;
 
 	/* Open the command for reading. */
-	f = popen("getprop | grep '\\.dns*.\\]\\:'", "r");
+	f = popen("getprop | grep '\\..*\\.dns[0-9]\\]:'", "r");
 	if (!f)
 		return errno;
 
--- a/src/dns/ns.c
+++ b/src/dns/ns.c
@@ -13,6 +13,7 @@
 #include "dns.h"
 #ifdef __ANDROID__
 #include <sys/system_properties.h>
+#include <stdlib.h>
 #endif
 
 
@@ -22,7 +23,7 @@
 
 
 static int parse_resolv_conf(char *domain, size_t dsize,
-			     struct sa *srvv, uint32_t *n)
+				 struct sa *srvv, uint32_t *n)
 {
 	FILE *f;
 	struct pl dom = pl_null;
@@ -62,7 +63,7 @@ static int parse_resolv_conf(char *domain, size_t dsize,
 
 		/* Use the first entry */
 		if (i < *n && 0 == re_regex(line, len, "nameserver [^\n]+",
-					    &srv)) {
+						&srv)) {
 			err = sa_set(&srvv[i], &srv, DNS_PORT);
 			if (err) {
 				DEBUG_WARNING("sa_set: %r (%m)\n", &srv, err);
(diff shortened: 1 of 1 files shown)


alfredh:
hi @Hmart 

many thanks for your patch.

A question about the new method for fetching DNS Servers; do you think that this
solution should be used for _all_ versions of Android, or should we check the version
and then use the old solution for <8.0 and the new solution for >= 8.0 ?

I tested your patch on an Android 6.0 phone. It seems to work but the
nameserver is added twice:

```
 DNS Servers from System: (2)
   0: 10.0.0.1:53
   1: 10.0.0.1:53
```

```
shell@j5xnlte:/data/local/tmp $ getprop | grep dns                             
[dhcp.wlan0.dns1]: [10.0.0.1]
[dhcp.wlan0.dns2]: []
[dhcp.wlan0.dns3]: []
[dhcp.wlan0.dns4]: []
[net.dns1]: [10.0.0.1]
```

@all it would be really helpful if you can help to test this patch
on your Android phone/emulator :)


Alfred

Hmart:
Hi @alfredh 

It seems that the nameserver is indeed added twice on android version < 8 since the net.dns properties are still accessible. 

I'll try to sort this issue out without adding the version check.

Hannes

Hmart:
The latest patch solves the issue with duplicated dns servers, however it seems like for some Android versions < 8.0 the other dns properties are not present and therefore no dns servers will be added. 

I'll add in a fallback, checking for the old properties aswell.

src/dns/ns.c
```diff
@@ -22,7 +23,7 @@
 
 
 static int parse_resolv_conf(char *domain, size_t dsize,
-			     struct sa *srvv, uint32_t *n)
+				 struct sa *srvv, uint32_t *n)
```
alfredh: this is an unrelated change, please revert it

src/dns/ns.c
```diff
@@ -62,7 +63,7 @@ static int parse_resolv_conf(char *domain, size_t dsize,
 
 		/* Use the first entry */
 		if (i < *n && 0 == re_regex(line, len, "nameserver [^\n]+",
-					    &srv)) {
+						&srv)) {
```
alfredh: this is an unrelated change, please revert it

alfredh:
I tested on an Android Oreo 8.0 emulator and got this:

```
$ getprop | grep dns                                                                               
[net.eth0.dns1]: [10.0.2.3]
[net.eth0.dns2]: []
[net.eth0.dns3]: []
[net.eth0.dns4]: []
```

is this what you expect?

apart from a couple of whitespace fixes, I think the patch is ready for merge.

Hmart:
Hi @alfredh  

The patch works as intended for all emulated devices I tested, however after testing the patch on several real devices updated to Android 8.0 it seems like the dns properties that are present in the emulated devices are removed all together. 

The patch is therefore not viable for at least the devices I tested today
Sony Xperia XZ F8331
Samsung s8

It seems the only viable fix for this problem is to set the dns server to a static ip as you suggested.
I'd suggest checking the android version and fallback to a static dns if the version > 8.0.

alfredh:
hi @Hmart 

thanks for testing again. It is too bad that it was not working on real devices.

the `dns_srv_get` function is supposed to be a quick way for the application
the fetch DNS nameservers from the system. If the application cannot use it, then
the application must implement its own methods or use static DNS servers.

1. adding a hardcoded DNS nameserver in the `libre` library is not an option.
  the library is generic and should not know about any special servers etc.
  also the library must work for IPv6 so then we have to add both an IPv4 and IPv6
  nameserver. This logic must be done in the application.

2. I would like to avoid adding Java/JNI code to libre. This looks like a very heavy solution
  just for fetching a DNS nameserver. Again, this code should be done in the application.

in any case, many thanks for trying to fix this.

Alfred

alfredh:
here is the output from another 8.0 device:

```
[ro.build.version.ota]: [OnePlus3Oxygen_16.A.62_GLO_062_1712272249]
[ro.build.version.release]: [8.0.0]
[ro.build.version.sdk]: [26]

getprop | grep dns
[init.svc.mdnsd]: [stopped]
```

i.e. no DNS nameservers from getprop. So that confirms it.

Hmart:
The fix does not work for most Android devices therefore I'll close this PR.

I'll reopen it if I find an alternative native solution to the problem.

Issue #140, opened 2018-05-18
Issues with includes


TheSil:
First of all, this library is great, I am using it to get familiar with RTP/RTCP protocols. During that, I have run into couple minor issues related to including re_* header files, so I thought it would be worth sharing:

1. The standard types `uint32_t`, etc, conflicts with `stdint.h`, which does not seem to be reflected anywhere in the library. Perhaps adding something like `HAVE_STDINT_H` would help?.

2. Again a bit related, for C++ code the `bool`, `true`, `false` and `inline `keywords are all redefined, but that does not seem to be a good idea (and some compilers such as Visual Studio 2017 evaluate this as an error).

3. (just a suggestion) The library relies heavily on importing just "re.h" (otherwise there are issues like missing include guards, etc), but since other includes are available as well, you can be sure people will include them. It might be worth considering either moving them out of API, or making them safe to include separately...

Apart from these I was able to use the library and it works well! (Well for now I have only simple RTP requirements, will be using it more thoroughly later).

alfredh:
hi,

your applications Makefile should include re.mk to get the same CFLAGS etc as re,
see the redemo project:

https://github.com/creytiv/redemo/blob/master/Makefile#L15

it is recommended that the application only includes `<re.h>`, see redemo:

https://github.com/creytiv/redemo/blob/master/src/sip_ua/sip_ua.c#L21

the redefinition of types is only a fallback solution, in case inttypes.h is not available.
make sure `HAVE_INTTYPES_H` is defined in your application.

either in your makefile or in your code:

```
#define HAVE_INTTYPES_H 1
#include <re.h>
```

Alfred

alfredh:
Would you like to submit a Pull Request for that change (inline and C++) ?

It would be great to test the change with different MS compilers.
(I dont have access to Windows machines).

TheSil:
Created the pull request, I have tested on latest Visual Studio 2017, but from what I remember, we have used C++ inline keyword for years (even in Visual Studio 6 which is like from 1998), and it worked fine, and for C it now works like before.

alfredh:
fixed in commit f9224f3edd2d1d5834af6b127fad5da6f369f639

Thanks @TheSil  !

PR #143, opened 2018-05-26
Inline C++ issue fix for MSVC

TheSil:
See https://github.com/creytiv/re/issues/140. Tested on Visual Studio 2017, would be great if someone can test for older VC versions. This article https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2013/z8y1yy88(v=vs.120) indicates that even in old version, the inline was available for C++, only issue is with C code, because of pure historical support of C99. By the way this has changed and it now supports inline as well, so the commit could be improved by checking for specific version of _MSC_VER < X (I don't know minimal possible X though), but it is not needed.

### DIFF
--- a/include/re_types.h
+++ b/include/re_types.h
@@ -93,7 +93,9 @@ typedef bool _Bool;
 
 /* Needed for MS compiler */
 #ifdef _MSC_VER
+#ifndef __cplusplus
 #define inline _inline
+#endif
 #endif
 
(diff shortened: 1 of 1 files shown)


alfredh:
many thanks for the patch. the proposed change looks fine to me.

@richaas if you agree, please review and merge to master.

PR #146, opened 2018-06-17
sip: get local TCP address in establish handler

alfredh:
after calling tcp_connect(), the address returned by
tcp_conn_local_get() is 0.0.0.0 on some platforms (Windows).

the local address of the TCP-connection is not available
until the TCP connection is fully established, so fetch the
address again in the established handler.

Reference: https://github.com/alfredh/baresip/issues/431

### DIFF
--- a/src/sip/transp.c
+++ b/src/sip/transp.c
@@ -459,6 +459,8 @@ static void tcp_estab_handler(void *arg)
 	struct le *le;
 	int err;
 
+	tcp_conn_local_get(conn->tc, &conn->laddr);
+
 	conn->established = true;
 
 	le = list_head(&conn->ql);
--- a/src/sip/transp.c
+++ b/src/sip/transp.c
@@ -459,7 +459,9 @@ static void tcp_estab_handler(void *arg)
 	struct le *le;
 	int err;
 
+#ifdef WIN32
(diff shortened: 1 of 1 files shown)


alfredh:
additional info:

the problem was reported on Windows 7 that the Contact header
contained the address 0.0.0.0

this address comes from this code in src/sip/transp.c:

```
	err = tcp_connect(&conn->tc, dst, tcp_estab_handler, tcp_recv_handler,
			  tcp_close_handler, conn);
	if (err)
		goto out;

	err = tcp_conn_local_get(conn->tc, &conn->laddr);
	if (err)
		goto out;
```

This code works fine on e.g. OSX and Linux. But on windows the
address in `conn->laddr` is 0.0.0.0 -- because the TCP connection
is not established at this point.

The windows documentation says:

"A Windows Sockets application must not assume that the address will be specified unless the socket is connected."

https://msdn.microsoft.com/en-us/library/windows/desktop/ms738543(v=vs.85).aspx


The question is also, should we update `tcp_conn_local_get` to check
the connection state ?


Alfred

richaas:
I suggest we wrap this in #ifdef WIN32

alfredh:
okay, I added a WIN32 check.

please merge to master if okay.

PR #147, opened 2018-06-18
jbuf: zero out jbuf_stat on jbuf flush

cspiel1:
When a jitter buffer is flushed also the statistic data should be cleared. That
means set back to zero. Except the field n_flush should be incremented. This fixes ugly warnings like
"jbuf: get: seq=32373 too late"
when the stream is resumed.


alfredh:
hi @cspiel1 

many thanks for your patch. the code looks fine to me.

@richaas please review and merge to master.

PR #149, opened 2018-08-02
Put intermediate build files in BUILD directory

AlexanderplUs:
I've changed Makefile to put resulted libre.a, libre.so and libre.pc in output BUILD directory. It is very useful during cross-compilation for different architectures. And it is common for all libraries with Makefiles.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,8 +47,9 @@ MKDIR   := $(PREFIX)/share/re
 CFLAGS	+= -Iinclude
 
 MODMKS	:= $(patsubst %,src/%/mod.mk,$(MODULES))
-SHARED  := libre$(LIB_SUFFIX)
-STATIC	:= libre.a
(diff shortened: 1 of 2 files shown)


alfredh:
hi @AlexanderplUs 

many thanks.

could you please split this PR into 2 PRs ? one PR with adding the PKG_CONFIG
variable and one PR with moving the libs.

did you test that the `make install` target is still working ? could you also
test if `make deb` and `make rpm` still works ?

the path change will break some other project that uses libre,
for example baresip-android:

https://github.com/alfredh/baresip-android

so we also need to find a solution for that ..

AlexanderplUs:
Ok, I'll split PR in 2 PRs.

`make install`, `make deb` and `make rpm` works fine.
Path change may break only projects that doesn't use installed libraries which is dangerous anyway.

alfredh:
thanks @AlexanderplUs 

I agree that projects should not use direct path to libre.a :)

Issue #151, opened 2018-08-17
Importing sip trace capability into official repo


quentusrex:
Pulling commit https://github.com/traviscross/libre/commit/37e7252269c49af4c75d8d4f41ea164a8527154c

PR #152, opened 2018-08-17
Add ability to print SIP trace

quentusrex:
Resolves https://github.com/creytiv/re/issues/151


quentusrex:
@alfredh Just pinging here to refresh memory, as we worked on this patch a little while back.

alfredh:
thanks @quentusrex :)

yes I did not forget about it. we will see how to proceed with this one.

alfredh:
an alternative solution would be to expose the UDP/TCP sockets in the
SIP transport API.

then the application could use UDP/TCP helpers to fetch all outgoing/incoming
packets.

alfredh:
we have decided not to apply this patch.

we have a better solution for this, but we are not sure when it will be added.

for now you can keep the patch in your fork.

PR #155, opened 2018-08-30
tls: set verify methode for openSSL 1.1.0

wnetbal:
Without this change the CA will be read correct but
the CA will never be used for checks.

This pull request fixes issue 154

### DIFF
--- a/src/tls/openssl/tls.c
+++ b/src/tls/openssl/tls.c
@@ -182,6 +182,10 @@ int tls_alloc(struct tls **tlsp, enum tls_method method, const char *keyfile,
 			goto out;
 		}
 
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
+		SSL_CTX_set_verify(tls->ctx, SSL_VERIFY_PEER, NULL);
+#endif
+
 		r = SSL_CTX_use_PrivateKey_file(tls->ctx, keyfile,
 						SSL_FILETYPE_PEM);
 		if (r <= 0) {


alfredh:
hi @wnetbal 

can you please tell me how you are using the API and how I can test this?
i.e. are you using libre for TLS client or TLS server (or both) ?

in tls_alloc, the if (keyfile) check if normally for a server. but it can also be used
by a TLS client, if the TLS server requires a client cert.

we also have another function to enable the TLS server to ask for TLS cert
from the client:

```c
void tls_set_verify_client(struct tls *tls)
{
	if (!tls)
		return;

	SSL_CTX_set_verify_depth(tls->ctx, 0);
	SSL_CTX_set_verify(tls->ctx, SSL_VERIFY_PEER | SSL_VERIFY_CLIENT_ONCE,
			   verify_handler);
}
```

this function is used for example with DTLS-SRTP, where both the
DTLS client and server have a certificate, which is mutually authenticated.

wnetbal:
Alfred E. Heggestad <notifications@github.com> schrieb am So., 2. Sep.
2018, 14:57:

> hi @wnetbal <https://github.com/wnetbal>
>
> can you please tell me how you are using the API and how I can test this?
> i.e. are you using libre for TLS client or TLS server (or both) ?
>
I use baresip to connect to a sip server with transport=tls
The certificate was merged from a .crt, a .key and the ca.crt file to a
.pem file.
The name of the pem file I set to the sip_certificate value in the baresip
configuration.
For the server (asterik) I created the certificates with the same CA like I
used for the pem file.

in tls_alloc, the if (keyfile) check if normally for a server. but it can
> also be used
> by a TLS client, if the TLS server requires a client cert.
>
> we also have another function to enable the TLS server to ask for TLS cert
> from the client:
>
> void tls_set_verify_client(struct tls *tls)
> {
> 	if (!tls)
> 		return;
>
> 	SSL_CTX_set_verify_depth(tls->ctx, 0);
> 	SSL_CTX_set_verify(tls->ctx, SSL_VERIFY_PEER | SSL_VERIFY_CLIENT_ONCE,
> 			   verify_handler);
> }
>
> I have seen this function and tried it, but it didnt work in my test.
It looks like this function was ignored or use later.
I'm sorry that I can not tell you the exact line of code, but I am on
vacation right now an have no access to my data.

> this function is used for example with DTLS-SRTP, where both the
> DTLS client and server have a certificate, which is mutually authenticated.
>
I have seen this and use this modul as well and IT works great.

Kind regards
Wolfgang

alfredh:
hi @wnetbal 

are you back from your holidays ?

alfredh:
closing this one for now, please re-open if you have time to answer my questions ..

wnetbal:
Hi @alfredh 

yes I'm back from holidays.

I added function tls_set_verify_client() in baresip src/ua.c line 1179 right after the error handling of tls_alloc().
But this doesn't worked for me, only when I added the code to the tls_alloc() function it worked.

What I found out in between is that this fix is only the half solution because it looks like the CA verification always fails, when the CA is in the .pem file.

To fix this I extendend baresip to add the CA by calling tls_add_ca() after tls_alloc()
(see baresip pull request "Use extra CA file for TLS transport #543")

Kind regards
Wolfgang

src/tls/openssl/tls.c
```diff
@@ -182,6 +182,10 @@ int tls_alloc(struct tls **tlsp, enum tls_method method, const char *keyfile,
 			goto out;
 		}
 
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
+		SSL_CTX_set_verify(tls->ctx, SSL_VERIFY_PEER, NULL);
+#endif
+
```
alfredh: regarding this diff I have two comments:

1. one would think that the default behaviour for openssl clients is to
  verify the TLS Server ? If not it makes the whole certificate stuff useless.
  perhaps the code is needed, but do you have references to relevant docs ?

2. the code block is inside a block (`if (keyfile)`) that is only executed if the
  certificate parameter is specified. This is typically done for TLS servers, but
  I think your usecase was a TLS client ? perhaps the code block should be moved
  to a place that is always executed for both client/server ?

alfredh:
I will close this PR now.

I have tested with tls_peer_verify and it works fine with both
openssl 1.0.2 and 1.1.0

if you are still sure that this is a bug, please provide more details/evidence
about how it is used, certificates, errors etc.


Alfred

PR #156, opened 2018-09-07
sip: Avoid function sip_addr_decode to clear uri-parameters.

cspiel1:
The function sip_addr_decode has two branches. One for SIP addresses with angle brackets, like

Alice <sip:alice@atlanta.com>

and one for addresses without angle brackets, like

sip:alice@atlanta.com

But what if we need some URI parameters, like:

Alice \<sip:alice@atlanta.com;transport=tcp\>

or

sip:alice@atlanta.com;transport=tcp

Then the parameters are decoded to different places. In the first branch, they will be part of addr->auri. And addr->params is empty. Also addr->uri.params will contain the parameters.

In the second branch the address is split at the ';'. The addr->auri will NOT contain the parameters. Neither addr->uri.params.

RFC 3261: 
Page 222 tells me that transport=tcp is an uri-parameter. But section 20.10 specifies that if no angle brackets are present transport=tcp is a header parameter. This makes no sense for me.

The libre/baresip code tells me the truth. And I think this is that in call->peer_uri the parameters (e.g. transport=tcp) should be contained. See
function call_connect, where sip_addr_decode is used.

Currently calling "sip:alice@atlanta.com;transport=tcp" will not work correctly and UDP is used as transport protocol.

Regards,c.

### DIFF
--- a/src/sip/addr.c
+++ b/src/sip/addr.c
@@ -41,11 +41,13 @@ int sip_addr_decode(struct sip_addr *addr, const struct pl *pl)
 			addr->params.p = NULL;
 	}
 	else {
+		struct pl params;
 		memset(addr, 0, sizeof(*addr));
 
 		if (re_regex(pl->p, pl->l, "[^;]+[^]*",
-			     &addr->auri, &addr->params))
+			     &addr->auri, &params))
 			return EBADMSG;
+		addr->auri = *pl;
 	}
 
 	err = uri_decode(&addr->uri, &addr->auri);


richaas:
Hi @cspiel1, I belive the current code is correct and according to RFC:

> If no "<" and ">" are present, all parameters after the URI are header parameters, not URI parameters.

cspiel1:
Dear richaas,
so transport=tcp is an uri-parameter if specified inside angle brackets, but header parameter if the angle brackets are omitted?
For baresip transport=tcp have to be decoded into addr->uri.params. Right? So the patch is essential, or it has to be solved differently.
r,c.

richaas:
For SIP, transport is always an URI parameter.

alfredh:
hi @cspiel1 

this is a bug in baresip, please see here for more info:

https://github.com/alfredh/baresip/issues/532

cspiel1:
> For SIP, transport is always an URI parameter.

Well, the patch does exactly that.

alfredh:
hi @cspiel1 

we have fixed the issue in baresip now, calling with ;transport=tcp should work now.
can you please test git HEAD ?

also thanks for this patch, but I believe that it is not needed.



Alfred

cspiel1:
Hi Alfread,
I followed your discussion. Today I will try to understand, verify and test the git HEAD. What I can say for now is, that without a patch for sip_addr_decode this issue can't be solved. 

regards,chris

cspiel1:
Hi Alfred,
ok, the commit ec33a4bb3c08beb9491e8b2e73d9ebc25673ce0a changes something for the issue #532. But it is not perfect. As you mentioned the transport parameter can be twice:

/dial sip:x@192.168.110.65;transport=tcp
call: connecting to 'sip:x@192.168.110.65;transport=tcp;transport=udp'..

Here is the INVITE:
INVITE sip:x@192.168.110.65;transport=tcp;transport=udp SIP/2.0

The problem I mentioned is that sip_addr_decode is wrong. I added a unit test in:
https://github.com/cspiel1/retest.git

siptest: 25: sip addr cmp: ref='sip:alice@atlanta.com;transport=tcp;jackson=1' gen='<sip:alice@atlanta.com>;transport=tcp;jackson=1'
test: test_sip_addr: test failed (Invalid argument)

Every occurance of sip_addr_decode may result in a failure.

account, contact: Do you only want entries with angle brackets?

message: Failure in case of missing angle brackets.

reg: ?

regards,
chris.

alfredh:
hi @cspiel1 

this is a bug in baresip and not in libre. could you please create a new Issue for this
in the baresip project?

some of things that we should fix there:

* ua_connect should not append the ;transport=xxx parameter
* functions calling ua_connect should convert from a SIP address to a request uri


Alfred

PR #158, opened 2018-10-01
tls: rename input argument to tls_add_ca

alfredh:


alfredh:
the reason for this patch is that the file/path name is swapped.

here is the prototype for the OpenSSL function:

```
int SSL_CTX_load_verify_locations(SSL_CTX *ctx, const char *CAfile,
                                   const char *CApath);

```

i.e. since we use argument number 2 (CAfile) it might make sense to also
rename the argument in the libre wrapper.

@richaas please review and merge to master if you agree.

PR #159, opened 2018-10-08
Handle EPOLLHUP

lgrahl:
I've used `fd_listen` for a pipe a while ago and encountered that `EPOLLHUP` was not handled. This PR raises `FD_EXCEPT` in case the `EPOLLHUP` event is being catched.

I have since removed the code that used the pipe from my code but thought this might still be a useful addition.

### DIFF
--- a/src/main/main.c
+++ b/src/main/main.c
@@ -784,7 +784,7 @@ static int fd_poll(struct re *re)
 				flags |= FD_READ;
 			if (re->events[i].events & EPOLLOUT)
 				flags |= FD_WRITE;
-			if (re->events[i].events & EPOLLERR)
+			if (re->events[i].events & (EPOLLERR|EPOLLHUP))
 				flags |= FD_EXCEPT;
 
 			if (!flags) {


alfredh:
hi @lgrahl 

many thanks for this patch. do you have some sample testcode that I can use
to test your patch ?


Alfred

lgrahl:
I'll see what I can dig up from my repo. :)

alfredh:
hi @lgrahl 

did you find the testcode for the pipe ?

I tried to test it locally with mqueue/pipe but I did not get the EPOLLHUP event on close.

here are some additional comments:

1. did you test with other I/O methods such as `select` and `poll` ?
  from the re API perspective the behaviour should be as similar as possible.

2. did you test it on different Linux distributions and kernels ?
  since this is a low-level change we need to test that it does not break anything else.

3. do you know if other Async I/O libraries handle EPOLLHUP ?


from the epoll man page:

```
       EPOLLHUP
              Hang up happened on the associated file descriptor.  epoll_wait(2) will  always  wait  for  this
              event; it is not necessary to set it in events.  Note that when reading from a channel such as a
              pipe or a stream socket, this event merely indicates that the peer closed its end of  the  chan‐
              nel.   Subsequent  reads from the channel will return 0 (end of file) only after all outstanding
              data in the channel has been consumed.
```

lgrahl:
I haven't found time yet to dig into my code and come up with a test case. :slightly_frowning_face:

1. No but once we have a test, we can check it.
2. I only tested it on a couple of Linux kernels.
3. I briefly looked into libuv which [also handles it](https://github.com/libuv/libuv/blob/master/src/unix/posix-poll.c) but I haven't looked how it handles it.

alfredh:
the change is useful, but since it does not solve a known problem in libre,
I suggest we drop this one for now. the re main code is by now very mature
and running in many production systems. we should not make any changes there
unless it solves a critical bug.

lgrahl:
I finally managed to find some time and write a test case, see https://github.com/creytiv/retest/pull/8.

alfredh:
thanks @lgrahl for the test :)

I compiled it on Linux and OSX and tested it (with some modifications)

I assume that the test exit-criteria is to receive the `FD_EXCEPT` event ?

I tested with 4 different polling methods, showing different results:

* poll (Linux) -- got FD_EXCEPT
* select (Linux) -- no event
* epoll (Linux) -- got FD_EXCEPT with patched libre main (EPOLLHUP)
* kqueue (OSX) -- no event
* select (OSX) -- no event

lgrahl:
> I assume that the test exit-criteria is to receive the `FD_EXCEPT` event

To be honest I'm not exactly sure which event is supposed to be raised when a pipe (or an attached pseudo-terminal) is being closed. I can't find any documentation that describes the intended behaviour. However, `read` does describe that it will return `0` for EOF but I also encountered `-1` plus `errno = EIO`.

I just updated the test to be compatible with select and poll as well. Can you test it on OSX?

alfredh:
thanks! I tested again and it is all working good with all I/O methods.

src/main/main.c
```diff
@@ -786,6 +786,8 @@ static int fd_poll(struct re *re)
 				flags |= FD_WRITE;
 			if (re->events[i].events & EPOLLERR)
 				flags |= FD_EXCEPT;
+			if (re->events[i].events & EPOLLHUP)
+				flags |= FD_EXCEPT;
```
alfredh: this can be simplified by oring EPOLLHUP with EPOLLERR:

```
	if (re->events[i].events & (EPOLLERR|EPOLLHUP) )
		flags |= FD_EXCEPT;
```

alfredh:
this is how HUP is handled with poll:


```
			if (re->fds[fd].revents & (POLLERR|POLLHUP|POLLNVAL))
				flags |= FD_EXCEPT;
```


given that the test is now behaving identical on all platforms/poll-methods
I think we should go ahead and merge the fix to master.

@richaas please merge to master if you agree

PR #161, opened 2018-10-12
Make headroom of an outgoing DTLS packet configurable

lgrahl:
This PR allows to configure the *headroom* of an outgoing DTLS packet.

Adds the functions `dtls_headroom` and `dtls_set_headroom`.

### DIFF
--- a/include/re_tls.h
+++ b/include/re_tls.h
@@ -77,6 +77,8 @@ int dtls_listen(struct dtls_sock **sockp, const struct sa *laddr,
 		dtls_conn_h *connh, void *arg);
 struct udp_sock *dtls_udp_sock(struct dtls_sock *sock);
 void dtls_set_mtu(struct dtls_sock *sock, size_t mtu);
+size_t dtls_headroom(struct dtls_sock *sock);
+void dtls_set_headroom(struct dtls_sock *sock, size_t headroom);
 int dtls_connect(struct tls_conn **ptc, struct tls *tls,
 		 struct dtls_sock *sock, const struct sa *peer,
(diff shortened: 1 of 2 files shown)


alfredh:
@lgrahl thanks for your patch.

I assume that you want to control if outgoing DTLS packets are sent via TURN or not ?

The TURN client has some advanced logic for deciding if the outgoing
packet should be sent directly or via TURN. The logic includes checking the number
of bytes available as headroom before pos (4 bytes for channel) and if there exist
a created channel/permission or not.

since the UDP socket is shared between multiple protocol layers, we cannot
set this flag (TURN or not TURN) on the socket level. I guess it has to be set
on a per-packet level.

I think it would be nice to re-design this logic a bit and use some kind of
explicit indication instead. But I am not sure exactly how that would look like :)

is my assumption correct, or do you have a different usecase ?
do you prefer to send DTLS packets always over TURN Channel or SendIndication ?

alfredh:
perhaps a simpler solution is to increase the headroom to 36.

this will have enough space for either Channel header or Send-Indication header.

lgrahl:
There may be other use cases and the changes of this PR are quite trivial, so I would prefer it to be configurable. Also, if I'm counting correctly, `36` will be insufficient for IPv6.

weili-jiang:
I hit the TURN headroom issue doing my own custom UDP application payload as well. I eventually stumbled upon the magic headroom numbers in the BFCP implementation but some way of setting it explicitly would be nice.

alfredh:
we have decided to not merge this one to master.

@lgrahl for now you can maintain this patch in your fork.

PR #166, opened 2018-10-30
rtmp: handle both IPv4 and IPv6 uris

alfredh:
this PR adds support for IPv6 uris.

after this PR the following URI types should be supported:

* IPv4 address
* IPv6 address
* FQDN domain name

Example URI:

 * rtmp://example.com/vod/mp4:sample.mp4
 * rtmp://127.0.0.1/vod/mp4:sample.mp4
 * rtmp://[::1]/vod/mp4:sample.mp4

The IPv6 address must be encapsulated in square brackets `[]` similar
to other protocols such as SIP.

### DIFF
--- a/src/rtmp/conn.c
+++ b/src/rtmp/conn.c
@@ -792,9 +792,8 @@ int rtmp_connect(struct rtmp_conn **connp, struct dnsc *dnsc, const char *uri,
 		return EINVAL;
 
 	if (re_regex(uri, strlen(uri), "rtmp://[^/]+/[^/]+/[^]+",
-		     &pl_hostport, &pl_app, &pl_stream)) {
+		     &pl_hostport, &pl_app, &pl_stream))
 		return EINVAL;
-	}
 
 	if (uri_decode_hostport(&pl_hostport, &pl_host, &pl_port))
 		return EINVAL;
--- a/src/rtmp/conn.c
+++ b/src/rtmp/conn.c
@@ -757,6 +757,19 @@ static void query_handler(int err, const struct dnshdr *hdr, struct list *ansl,
 }
 
 
+static int decode_hostport(const struct pl *hp, struct pl *host,
+			   struct pl *port)
+{
+	/* Try IPv6 first */
+	if (!re_regex(hp->p, hp->l, "\\[[0-9a-f:]+\\][:]*[0-9]*",
+		      host, NULL, port))
+		return 0;
+
+	/* Then non-IPv6 host */
+	return re_regex(hp->p, hp->l, "[^:]+[:]*[0-9]*", host, NULL, port);
+}
+
+
 /**
  * Connect to an RTMP server
  *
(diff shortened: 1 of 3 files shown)


alfredh:
can you give an example of such an URI with Scope ID and how we can transfer
it to the `struct sa` socket address ?

lgrahl:
Disclaimer: This isn't limited to RTMP. I have encountered this in ICE as well, see https://github.com/rawrtc/rawrtc/issues/62.

A URL example: `http://[fe80::1%eth0]:4242/foo/bar`.
To transfer this into `struct sa`, `struct sockaddr_in6` provides a member called `sin6_scope_id` on Linux >= 2.4. FreeBSD provides that, too. And I guess the rest of the OSes will have some concept for that as well.

src/rtmp/conn.c
```diff
@@ -788,9 +803,14 @@ int rtmp_connect(struct rtmp_conn **connp, struct dnsc *dnsc, const char *uri,
...
 		return EINVAL;
 
-	if (re_regex(uri, strlen(uri), "rtmp://[^:/]+[:]*[0-9]*/[^/]+/[^]+",
-		     &pl_host, NULL, &pl_port, &pl_app, &pl_stream))
+	if (re_regex(uri, strlen(uri), "rtmp://[^/]+/[^/]+/[^]+",
+		     &pl_hostport, &pl_app, &pl_stream)) {
 		return EINVAL;
+	}
+
+	err = decode_hostport(&pl_hostport, &pl_host, &pl_port);
```
richaas: Suggest we do like this to match existing return code for invalid URIs:
```suggestion
	if (decode_hostport(&pl_hostport, &pl_host, &pl_port))
		return EINVAL;
```

richaas:
To avoid duplicated code, I suggest we make this function in src/uri/uri.c public:
```
-static int decode_hostport(...)
+int uri_decode_hostport(...)
```

alfredh:
thanks for review comments, code is updated.

@lgrahl there is no other code in libre that handles the link-local scope ID for IPv6,
so I dont think that RTMP should be the first one. converting from interface name
to the numerical ID sounds complex yes :)

src/rtmp/conn.c
```diff
@@ -788,8 +791,12 @@ int rtmp_connect(struct rtmp_conn **connp, struct dnsc *dnsc, const char *uri,
 	if (!connp || !uri)
 		return EINVAL;
 
-	if (re_regex(uri, strlen(uri), "rtmp://[^:/]+[:]*[0-9]*/[^/]+/[^]+",
-		     &pl_host, NULL, &pl_port, &pl_app, &pl_stream))
+	if (re_regex(uri, strlen(uri), "rtmp://[^/]+/[^/]+/[^]+",
```
richaas: brackets not needed

src/uri/uri.c
```diff
@@ -74,10 +74,14 @@ int uri_encode(struct re_printf *pf, const struct uri *uri)
...
+ * @param port     Decoded port portion
+ *
  * @return 0 if success, otherwise errorcode
  */
-static int decode_hostport(const struct pl *hostport, struct pl *host,
-			   struct pl *port)
+int uri_decode_hostport(const struct pl *hostport, struct pl *host,
+			struct pl *port)
 {
 	/* Try IPv6 first */
```
richaas: Now that this is public, we should check input arguments (at least hostport).

alfredh:
code updated with review comments.

PR #173, opened 2018-12-07
rtmp: add rtmp_meta() to send metadata on stream

alfredh:
some RTMP servers expect to receive this message.
keep the content minimal for now.

### DIFF
--- a/src/rtmp/stream.c
+++ b/src/rtmp/stream.c
@@ -210,6 +210,21 @@ int rtmp_publish(struct rtmp_stream *strm, const char *name)
 }
 
 
+int rtmp_meta(struct rtmp_stream *strm)
+{
+	if (!strm)
+		return EINVAL;
+
+	return rtmp_amf_data((struct rtmp_conn *)strm->conn, strm->stream_id,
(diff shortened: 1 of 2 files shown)


richaas:
Proposing to alter the prototype for `rtmp_amf_data()` and add `const` for `struct rtmp_conn *`
to avoid having to case from stream.c:
```
int rtmp_amf_data(const struct rtmp_conn *conn, uint32_t stream_id,
		  const char *command, unsigned body_propc, ...)
```

src/rtmp/stream.c
```diff
@@ -210,6 +210,28 @@ int rtmp_publish(struct rtmp_stream *strm, const char *name)
...
+ *
+ * @return 0 if success, otherwise errorcode
+ */
+int rtmp_meta(struct rtmp_stream *strm)
+{
+	if (!strm)
+		return EINVAL;
+
+	return rtmp_amf_data(strm->conn, strm->stream_id,
+			     "@setDataFrame",
```
richaas: Suggest moving this one line up

PR #177, opened 2018-12-30
modify mem_deref

cloud-jxy:
I code the patch to avoid crash in my project.

I use libre to develop other project. I found libre occour crash on mem_deref. The value of struct mem object is error. In my project , I have found the evidence that a pointer may mutli-free, especially in multi thread. Memory lack is another reason that I hasn't found the evidence to prove or disprove it.


alfredh:
the error is most likely in your program.

if you are accessing the same object from 2 or more different threads,
you need to use pthread mutex or `struct lock` to protect the object.

PR #178, opened 2019-01-10
rpm: add missing file to %install section

stv0g:


alfredh:
thanks for the patch.

I would like to suggest that we remove the `rpm` folder and the spec file.
The RPM can still be built outside the tree.

stv0g:
@alfredh 

> I would like to suggest that we remove the rpm folder and the spec file. The RPM can still be built outside the tree.

I agree. Where would you place it? Creating a new packaging repo is a bit overkill in my opinion.

alfredh:
we have agreed to keep the rpm for now.

@richaas please merge to master if you agree

PR #183, opened 2019-01-28
rtmp: add extended timestamp for type 3

alfredh:
The extended timestamp should also be present for Type 3 packets.

### DIFF
--- a/src/rtmp/chunk.c
+++ b/src/rtmp/chunk.c
@@ -67,6 +67,10 @@ int rtmp_chunker(unsigned format, uint32_t chunk_id,
 		chunk_sz = min(len, max_chunk_sz);
 
 		err  = rtmp_header_encode(mb, &hdr);
+
+		if (timestamp >= 0xffffff)
+			err |= mbuf_write_u32(mb, htonl(timestamp));
+
 		err |= mbuf_write_mem(mb, payload, chunk_sz);
 		if (err)
 			goto out;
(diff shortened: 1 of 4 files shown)


richaas:
A question regarding dechunking and extended timestamp:
Consider the scenario in which chunk->hdr.timestamp has been incremented >= 0x00ffffff by receiving multiple type 1, 2 or 3 messages and the first type 0 message that created the initial
dechunker chunk state had a timestamp < 0x00ffffff. Should we still expect the type 3 message to contain an extended timestamp field in this case?

alfredh:
as far as I can understand, we have to look at the most recent type 0,1,2 packet
and see if the timestamp exceeds 0x00ffffff.

from the RTMP Specification:

```
5.3.1.3.  Extended Timestamp

   The Extended Timestamp field is used to encode timestamps or
   timestamp deltas that are greater than 16777215 (0xFFFFFF); that is,
   for timestamps or timestamp deltas that don’t fit in the 24 bit
   fields of Type 0, 1, or 2 chunks.  This field encodes the complete
   32-bit timestamp or timestamp delta.  The presence of this field is
   indicated by setting the timestamp field of a Type 0 chunk, or the
   timestamp delta field of a Type 1 or 2 chunk, to 16777215 (0xFFFFFF).
   This field is present in Type 3 chunks when the most recent Type 0,
   1, or 2 chunk for the same chunk stream ID indicated the presence of
   an extended timestamp field.
```

I have updated the patch with some improvements.

richaas:
I guess we also need to add support for encoding and decoding extended timestamp deltas for type 1 and 2 headers.

PR #185, opened 2019-02-02
Fix libdir in pkgconfig

Lastique:
The library directory can be specified by user in LIBDIR but it was not reflected in libre.pc. This commit fixes this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ endif
 libre.pc:
 	@echo 'prefix='$(PREFIX) > libre.pc
 	@echo 'exec_prefix=$${prefix}' >> libre.pc
-	@echo 'libdir=$${prefix}/lib' >> libre.pc
+	@echo 'libdir=$(LIBDIR)' >> libre.pc
 	@echo 'includedir=$${prefix}/include/re' >> libre.pc
 	@echo '' >> libre.pc
 	@echo 'Name: libre' >> libre.pc


alfredh:
I am not sure if this is needed. 

`PREFIX` is already configurable.

adding @lgrahl for extra pair of eyes ...

Lastique:
This is needed because libraries are not always in `lib`. On Debian and derivatives, libraries are normally in `lib/<multilib-tag>`, e.g. `lib/x86_64-linux-gnu`. You already have `LIBDIR` to configure that in the build process, but the pkg-config file doesn't follow that.

alfredh:
@Lastique thanks for the description :)

I checked the opus package in Debian 9 and the pc file looks like this:

```
# Opus codec reference implementation pkg-config file

prefix=/usr
exec_prefix=${prefix}
libdir=${prefix}/lib/x86_64-linux-gnu
includedir=${prefix}/include

Name: Opus
Description: Opus IETF audio codec (floating-point build)
URL: https://opus-codec.org/
Version: 1.2~alpha2
Requires:
Conflicts:
Libs: -L${libdir} -lopus
Libs.private: -lm
Cflags: -I${includedir}/opus
```

i.e. libdir has the multi-arch tag as you said.

should be okay to merge this on to master.

@richaas please merge if you agree ...

PR #187, opened 2019-02-06
Updating odict, websock and re.mk doxygen

metadings:
This PR is adding a function `odict_new`, to allow to create an object in one line of code.

This is also adding `struct websock_conn *` to the arguments of `websock_recv_h` and `websock_close_h`.

In the makefile re.mk I've updated doxygen, to work without perl and graphwiz-dot. I didn't call `doxygen -u mk/Doxyfile` because this seems to be a huge change... What do you think? 

Updating .travis.yml: Doesn't bash just read = instead of == ?

About the formatting issues: Don't you want to rethink these expects? For example,
a line 120 bytes, with a line count of at least max 2^16 == 65535 lines * 121 = 7929735 bytes ~ 8 MB,
or a line 96 bytes, with a line count of at least max 2^16 lines * 6356895 ~ 6 MB?

Actually I'm warming up with using git and github... Is this PR okay, or should I split those changes somehow?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,10 @@ install:
     - wget "https://github.com/alfredh/pytools/raw/master/ccheck.py"
 
 script: 
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+  - if [[ "$TRAVIS_OS_NAME" = "linux" ]]; then
         make EXTRA_CFLAGS=-Werror CCACHE=;
     fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+  - if [[ "$TRAVIS_OS_NAME" = "osx" ]]; then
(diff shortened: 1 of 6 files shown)


metadings:
I'm sorry Alfred Heggestad, I didn't want to burst in here !

metadings:
I'm sorry I burst here in like this 
Please look into the first three commits.

I've created a new 
```javascript
	struct odict *odict_new(uint32_t hash_size);
```
which you can use to create a new odict:
```javascript
	struct odict * answer0 = odict_new(32);
```
I've also added `struct websock_conn *` to `websock_recv_h` and to `websock_close_h`, because it's natural to have the object for which you're calling these handlers, as they now look like
```javascript
static void do_WebSocket_receive(struct websock_conn * connection, struct websock_hdr * hdr, struct mbuf * mb, void * arg) { /**/ }
static void do_WebSocket_close(struct websock_conn * connection, int err, void * arg) { /**/ }
```
Now look, I don't know how to remove these commits. I'm going to close this PR and refine it. G

PR #188, opened 2019-02-10
main: include all openssl cleanup recommendations

weili-jiang:
Fixes leaks detected by `valgrind` from `libre_init` even after `libre_close` is called.
The original clean-up code was missing much of the recommended calls listed in https://wiki.openssl.org/index.php/Library_Initialization#Cleanup

### DIFF
--- a/src/main/openssl.c
+++ b/src/main/openssl.c
@@ -158,8 +158,19 @@ int openssl_init(void)
 
 void openssl_close(void)
 {
+	// https://wiki.openssl.org/index.php/Library_Initialization#Cleanup
+	FIPS_mode_set(0);
+	EVP_PBE_cleanup();
+	EVP_cleanup();
+	CRYPTO_cleanup_all_ex_data();
 	ERR_free_strings();
-#if defined (HAVE_PTHREAD) && (OPENSSL_VERSION_NUMBER < 0x10100000L)
+	SSL_COMP_free_compression_methods();
+#if defined (HAVE_PTHREAD)
+#if (OPENSSL_VERSION_NUMBER < 0x10100000L)
+	ERR_remove_thread_state(NULL);
 	lockv = mem_deref(lockv);
+#elif (OPENSSL_VERSION_NUMBER < 0x10000000L)
+	ERR_remove_state(0);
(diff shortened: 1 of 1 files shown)


alfredh:
I am not sure if this is needed.

I have been running libre + openssl with valgrind and got zero warnings/errors.
this was tested on Linux.

can you please provide a full valgrind log and exact steps to reproduce,
including name/version of platform/os.

weili-jiang:
Oops, I see that some of the calls I added are OpenSSL version dependent and are failing the CI... some extra version guards are required I guess. I seem to be using OpenSSL 1.0.2n-1ubuntu5.2 on Ubuntu 18.04.2 LTS.

Compiled with `gcc main.c libre.a -Iinclude -lpthread -lcrypto -lssl -o main` from after compiling re:
[main.txt](https://github.com/creytiv/re/files/2860190/main.txt)

`valgrind --leak-check=full --show-leak-kinds=all --log-file=foo ./main`
[valgrind.log](https://github.com/creytiv/re/files/2860189/valgrind.log)

I think using OpenSSL 1.0 is the problem. If I switch to OpenSSL 1.1.0g-2ubuntu4.3 on the same OS it seems to be fine. So I can get around my immediate problem by switching to OpenSSL 1.1, but there is still a leak issue on OpenSSL 1.0.

alfredh:
I justed tested with Ubuntu 18.04 and OpenSSL 1.1.0 and there are zero valgrind warnings.

Is there any reason why you have to use an old version of OpenSSL
on Ubuntu when there is a new one available ?

The default openssl version on Ubuntu 18.04 is OpenSSL 1.1.0g

I have tested libre+openssl with valgrind on many different platforms,
and all of them gives zero valgrind warnings.

weili-jiang:
I can switch to OpenSSL 1.1.0, that's fine. I will close this PR since I can't be bothered trying to get the clean-up calls to compile against all versions of OpenSSL.

Note that in the README under "Supported versions of OpenSSL", "OpenSSL version 1.0.2" is listed, so if anybody is still using that they will have to deal with this.

Thanks

PR #190, opened 2019-02-13
rtmp: add server ping

alfredh:
The RTMP Server should send periodic PING to the Client.

* expose the control handler also for rtmp connection

* send Ping Request/response event to application

a different solution is to send periodic PING from within libre (RTMP module).


alfredh:
ping was implemented in the RTMP server. no need for changes in libre.

PR #197, opened 2019-04-08
dns: fix dname decode buffer checks

richaas:


alfredh:
Thanks to Konrad Beckmann who reported the bug!

PR #200, opened 2019-04-13
re_odict and re_websock

metadings:
Now I do want to make changes to your project, lets see :-)

I do want to add odict_new and odict_new_size.

I also do want to add `struct websock_conn *that` to instance handlers estab, close and recv.

### DIFF
--- a/include/re_websock.h
+++ b/include/re_websock.h
@@ -48,7 +48,7 @@ struct websock_conn;
 
 typedef void (websock_estab_h)(struct websock_conn * conn, void *arg);
 typedef void (websock_recv_h)(struct websock_conn * conn,
-    const struct websock_hdr *hdr, struct mbuf *mb, void *arg);
+	const struct websock_hdr *hdr, struct mbuf *mb, void *arg);
 typedef void (websock_close_h)(struct websock_conn*conn, int err, void*arg);
 
 
--- a/include/re_websock.h
+++ b/include/re_websock.h
@@ -46,10 +46,10 @@ struct websock_hdr {
 struct websock;
 struct websock_conn;
 
-typedef void (websock_estab_h)(void *arg);
-typedef void (websock_recv_h)(const struct websock_hdr *hdr, struct mbuf *mb,
-			      void *arg);
-typedef void (websock_close_h)(int err, void *arg);
+typedef void (websock_estab_h)(struct websock_conn * conn, void *arg);
+typedef void (websock_recv_h)(struct websock_conn * conn,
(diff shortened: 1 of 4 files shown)


metadings:
Listen, I do have a problem with your `struct mem`. Because just like dlmalloc you do pointer arithmetics to get this struct. The problem is, you can't ever use your structs really stackallocated, because 1. this struct is defined in a .c file, so you don't see this at all, 2. before stackallocating a struct you must define a struct mem, or you get a crash when doing mem_deref.

2. I do have a problem with your namings, of field names which are u, l, p and so say nothing, the same is with mem, pl, le, he, u, str and so on, and not Memory, Array or String, LinkedListElement, LinkedList and so on, which would make your code much more readable.

3. I don't know why you'd like your struct so much, because you do always use the pointers to your structs, in fact you must, since your struct mem isn't public. (Btw. mem_deref doesn't set the pointer to NULL).

Would you mind if I'm going to refactor this library?

I do want to use the "Public Domain" license, without "Copyright".

metadings:
Look: I do want to use the "Public Domain" license, without "Copyright".

```
Gemeinfrei. Public Domain.
2018 Uli Riehm
2010 Alfred E. Heggestad
```

Is that okay? I'm going to `git push metadings` a suggested refactored libreC... in a few days.

include/re_websock.h
```diff
@@ -46,10 +46,10 @@ struct websock_hdr {
...
 struct websock_conn;
 
-typedef void (websock_estab_h)(void *arg);
-typedef void (websock_recv_h)(const struct websock_hdr *hdr, struct mbuf *mb,
-			      void *arg);
-typedef void (websock_close_h)(int err, void *arg);
+typedef void (websock_estab_h)(struct websock_conn * conn, void *arg);
+typedef void (websock_recv_h)(struct websock_conn * conn,
+	const struct websock_hdr *hdr, struct mbuf *mb, void *arg);
+typedef void (websock_close_h)(struct websock_conn*conn, int err, void*arg);
```
alfredh: this change breaks the API and is not necessary.

the application should create an object that has a pointer to websock_conn.
the object should be passes in as "arg".

alfredh:
thanks for this patch.

the new functions introduced in the patch are not needed, and the changes
also breaks the websock API.

so I am sorry but we will not accept this patch.

regarding copyright, please read the copyright license in docs/COPYING.
the text clearly says what you can do and cannot do with the code.

include/re_websock.h
metadings: There would be a typed `struct websock_conn *that` instead of `void *arg`, in those events!

PR #203, opened 2019-04-24
no need to skip link-local IP, when scope id is set

juha-h:
Tested on Linux.

### DIFF
--- a/src/net/if.c
+++ b/src/net/if.c
@@ -3,6 +3,7 @@
  *
  * Copyright (C) 2010 Creytiv.com
  */
+#include <net/if.h>
 #include <re_types.h>
 #include <re_fmt.h>
 #include <re_mbuf.h>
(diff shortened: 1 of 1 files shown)


src/net/if.c
```diff
@@ -99,6 +100,15 @@ static bool if_getaddr_handler(const char *ifname,
...
 
+#ifdef HAVE_INET6
+	/* Assign IPv6 link local address scope id */
+	if (sa_is_linklocal(sa)) {
+		ife->ip->u.in6.sin6_scope_id = if_nametoindex(ifname);
+		DEBUG_WARNING("Set scope of %s to %u\n", ifname,
+			      ife->ip->u.in6.sin6_scope_id);
+	}
+
+#endif
```
alfredh: perhaps this should be moved to `net_if_getlinklocal` instead ?

alfredh:
the function that enumerates interfaces or returns the "default" address,
excludes link-local and loopback addresses by design. this has been done on purpose,
because most OS'es with IPv6 stack has a link-local address, but not global IPv6
connectivity.

some applications use this logic to decide if it should enable IPv6 or not
( e.g. if (have_ipv6) do_this() ).

I think the application should explicitly enable IPv6 link-local address
by using `net_if_getlinklocal`.

juha-h:
Alfred E. Heggestad writes:

> I think the application should explicitly enable IPv6 link-local address
> by using `net_if_getlinklocal`.

Fine, I'll close the pull request.

PR #210, opened 2019-05-08
Make able to set output handler instead of stdout

Encamy:
Hello!
There are some places where re_printf is used. Unfortunatelly, these messages can not be retarget to specific log handler. We are using re/rem/baresip as a library, so it will be awesome if application could control output.
Also, we are using patch https://github.com/creytiv/re/commit/49834b7afd0fe31660108271ef7657a34caa6a6d for windows and messages `$$$ NOTE - USING EXPERIMENTAL RE_MAIN $$$` and `fd_set updated` are spaming to console

This patch will not break any exist code, but will add possibility to set some specific log handler.

In this implementation method `re_vhprintf` can not be used instead of `re_vsnprintf` because in case of formatted string it will call log handler multiple times. For example message `re_printf("Test float = %f\n", 4.5f);` will convert into multiple function calls:
```
Test float = %f
4.5
<new line>
````

Usage example:
```c
static int my_re_log_handler(const char *p, size_t size, void *arg)
{
    // Some example logging
    some_external_logging(p, size, arg);
    return 0;
}

...
struct re_printf reLogHandler;
reLogHandler.vph = my_re_log_handler;
reLogHandler.arg = NULL;
re_set_output_handler(&reLogHandler);
```

### DIFF
--- a/include/re_fmt.h
+++ b/include/re_fmt.h
@@ -92,6 +92,8 @@ int re_printf(const char *fmt, ...);
 int re_snprintf(char *str, size_t size, const char *fmt, ...);
 int re_sdprintf(char **strp, const char *fmt, ...);
 
+int re_set_output_handler(struct re_printf *log_handler);
+
 
 /* Regular expressions */
 int re_regex(const char *ptr, size_t len, const char *expr, ...);
--- a/src/fmt/print.c
+++ b/src/fmt/print.c
@@ -43,6 +43,8 @@ static const char prfx_neg[]  = "-";
 static const char prfx_hex[]  = "0x";
 static const char str_nil[]  = "(nil)";
 
+static struct re_printf *re_log_handler = NULL;
+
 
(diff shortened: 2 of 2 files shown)


alfredh:
thanks for the patch, however we can't accept this in master.

the main reason is that it adds a global variable. we want libre to be fully re-entrant,
or at least as much as possible. adding global variables is going in the wrong direction ..

perhaps using `freopen` or something similar can work for you ?

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/freopen-wfreopen?view=vs-2019

btw, the other logging macros e.g. `DEBUG_NOTICE` has a way to set a callback
handler. please see `re_dbg.h`

Encamy:
Okay, thanks for comments
`freopen` will just redirect output from stdout to another file handler, it will not call some function
I will close this pr for now, maybe later I will made a better implementation

PR #212, opened 2019-05-24
natbd: remove module

alfredh:
This is a proposal to remove the NAT Behaviour Discovery module.

The code was used to try to detect the type of NAT that the device is behind,
however the technique is unstable and does not give same result every time.
The RFC is classified as Experimental.

If anyone is opposed to this change, please speak out.


alfredh:
I suggest we do this in the next major release.

PR #217, opened 2019-06-28
tls: don't close connection on SSL_ERROR_ZERO_RETURN

richaas:
If a TLS closure alert is observed before all received data is handed to the application, the TCP connection is closed and the data is lost. In the proposed fix we ignore the SSL_ERROR_ZERO_RETURN error, and leave the connection open until closed at the TCP level.

### DIFF
--- a/src/tls/openssl/tls_tcp.c
+++ b/src/tls/openssl/tls_tcp.c
@@ -306,12 +306,9 @@ static bool recv_handler(int *err, struct mbuf *mb, bool *estab, void *arg)
 
 			switch (ssl_err) {
 
+			case SSL_ERROR_ZERO_RETURN:
 			case SSL_ERROR_WANT_READ:
 				break;
-
-			case SSL_ERROR_ZERO_RETURN:
-				*err = ECONNRESET;
-				return true;
 
 			default:
 				*err = EPROTO;


alfredh:
I have tested the patch on the following platforms:

* OpenSSL 1.0.2 (OSX)
* OpenSSL 1.1.0j (Debian 9)
* OpenSSL 1.1.1c (OSX)

what should we do about UDP/DTLS, it has a similar logic:

```c
			case SSL_ERROR_ZERO_RETURN:
				conn_close(tc, ECONNRESET);
				return;
```

richaas:
Good work testing! I have written a test program that can reproduce this issue for TCP. I suggest we do the same for UDP and open a separate pull request if needed.

alfredh:
there is also a difference between TLS (TCP) and DTLS (UDP):

* TLS: one or more TLS records per TCP packet
* DTLS: exactly one DTLS record per UDP datagram

https://tools.ietf.org/html/rfc6347#section-4.1.1

"Each DTLS record MUST fit within a single datagram."

hence I dont think this problem exist with DTLS. the last application payload
to be received will be handled in the previous packet, before the Alert will close
the DTLS connection.

alfredh:
the issue can be reproduced with this patch to retest:

https://github.com/creytiv/retest/commit/9e1219ed97161ae38ce35b5481f2598728438c49

I have redone the test on selected platforms, see above.

1. with re master, this test fails.
2. with this patch, the test works.

I think it is safe to merge the patch to master.

PR #220, opened 2019-07-05
preliminary support for SIP/TLS cert verification

alfredh:
this patch needs a lot of testing :)

| SIP Server | Certificate | Status |
| ---- | ---- | ----|
| sip.antisip.com | Gandi Standard SSL CA 2 | ok |
| iptel.org | Comodo | fail |
| tutpro.com | Let's Encrypt | ok |


alfredh:
Intermediate certificates is not working with this patch.

one option is to use the http client and download the public intermediate cert
during TLS handshake, but this will add complexity.

I will come back with a better patch ..

(copy: https://gist.github.com/alfredh/d353de9645b4e78c24ed90e765406413)

PR #221, opened 2019-07-09
Honour interface preference

michaelmalonenz:
An interface or address preference can be set by clients.  However, this
is not always honoured - instead leaving it to the OS to choose the
interface on which to send the packets.  This is usually The Right Thing
to do as the OS is in the best place to choose the interface that will
most likely get the packet to the destination via the ip address / route
metric mechanisms.

However, in certain use cases where a clear preference has been
identified by the client, libre should be doing its best to use that as
the outgoing interface.  This can be achieved in two ways (that I know
of).

1) Set the socket option SO_BINDTODEVICE which requires net admin
privileges
or
2) Call bind() on the socket with the desired address which will set the
src address in the packets correctly, and it is then up to the OS for
which physical interface the packets are sent on, but with the src
address set, it is much easier to configure the OS to always send
those packets to a particular device.

I have avoided causing breaking changes to the public API, so I have added `*_hint` functions where applicable.  I am absolutely willing to modify this where necessary - the behaviour doesn't actually change if the `struct sa *` hint address passed in is NULL.

Mostly this is a PR to get a feeling for the appetite for this change.  It's kind of a special use case, so I don't know if it's actually something you are interested in maintaining.

### DIFF
--- a/include/re_dns.h
+++ b/include/re_dns.h
@@ -200,6 +200,7 @@ struct dnsc_conf {
 	uint32_t tcp_hash_size;
 	uint32_t conn_timeout;  /* in [ms] */
 	uint32_t idle_timeout;  /* in [ms] */
+	struct sa *laddr;
 };
 
 int  dnsc_alloc(struct dnsc **dcpp, const struct dnsc_conf *conf,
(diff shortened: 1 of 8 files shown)


alfredh:
many thanks for your patch

we will most likely not accept this patch, as we want the kernel's routing table
to take care of choosing the right network interface.

for example some VPN clients update the routing table to force all traffic
through the VPN.

if the application binds to a specific interface, and try to send packets to
a public host that is not routable from this interface, it will fail and the
end-user will be confused.

michaelmalonenz:
Ok no worries - I suspected that might be the case, but I thought I would share the patch in case you wanted it.

PR #222, opened 2019-07-12
use clock_gettime (if possible) to avoid problems with non-monotonic …

fwjavox:
This patch adds support for clock_gettime calls that use monotonic clocks that are not affected by leap seconds or summer/winter timezone changes.

### DIFF
--- a/src/tmr/tmr.c
+++ b/src/tmr/tmr.c
@@ -3,6 +3,10 @@
  *
  * Copyright (C) 2010 Creytiv.com
  */
+#define _POSIX_C_SOURCE 199309L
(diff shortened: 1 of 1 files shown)


alfredh:
thanks for your patch

we will most likely not accept this patch, please see here for discussion:

https://github.com/creytiv/re/pull/99

RomanED:
Hello! Sorry for off top, but I could not find information about my problem with libre. Can anyone tell me about the build libre for Windows x64? I have a problem and error something like "winsock error: fd=65", something wrong with dependencies? Or anything else?

PR #225, opened 2019-09-10
fixing crash when not using pthreads

srogers-plexsys:
If you aren't using pthreads, this file will cause baresip to crash. I have fixed this by replacing the pthreads mutex with the one used in re_lock.

### DIFF
--- a/src/dbg/dbg.c
+++ b/src/dbg/dbg.c
@@ -7,15 +7,12 @@
 #ifdef HAVE_UNISTD_H
 #include <unistd.h>
 #endif
-#ifdef HAVE_PTHREAD
-#include <stdlib.h>
-#include <pthread.h>
-#endif
 #include <time.h>
 #include <re_types.h>
 #include <re_fmt.h>
 #include <re_list.h>
 #include <re_tmr.h>
+#include <re_lock.h>
 
 
 #define DEBUG_MODULE "dbg"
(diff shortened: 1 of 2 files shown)


src/dbg/dbg.c
```diff
@@ -72,6 +62,9 @@ static inline void dbg_unlock(void)
  */
 void dbg_init(int level, enum dbg_flags flags)
 {
+	if (!dbg.mutex) {
+		lock_alloc(&dbg.mutex);
```
alfredh: return value of `lock_alloc` must be checked

alfredh:
can you please tell me which platform/os you are using and the details of the crash ?

srogers-plexsys:
Ah, yeah you are right. I will fix that. Thanks,

Steve Rogers

srogers-plexsys:
It happens in Windows, we are linking to a static library build of baresip. Windows version on my development machine is: 1803 Build 17134.1. We recently moved away from using pthreads in our main project after switching to std::thread in C++11. 

-Steve Rogers

alfredh:
are you logging from 2 different threads ?

I would like to see the full call-stack of the crash.

srogers-plexsys:
We are logging from multiple threads, yes. I cannot post a call stack from our software unfortunately. I could make a test app to induce it if you think it is worthwhile but I am not sure when I will have time to do that.

PR #226, opened 2019-09-16
do not fail authorization if it has succeed before

juha-h:
hack needed for servers that do not include stale=true in response

### DIFF
--- a/src/sip/auth.c
+++ b/src/sip/auth.c
@@ -161,7 +161,9 @@ static bool auth_handler(const struct sip_hdr *hdr, const struct sip_msg *msg,
 			goto out;
 	}
 	else {
-		if (!pl_isset(&ch.stale) || pl_strcasecmp(&ch.stale, "true")) {
+		/* error if first auth attempt fails */
+		if ((!pl_isset(&ch.stale) ||
+		     pl_strcasecmp(&ch.stale, "true")) && (realm->nc == 2)) {
 			err = EAUTH;
 			goto out;
 		}


richaas:
The SIP auth code relies on the nonce parameter to distinguish between two situations that both results in a 401/407 response; wrong credentials and stale nonce. Only the later case should trigger an immediate retry of the original request, otherwise it will result in a response/request loop (capped at 16 retries). As the credentials can change on the server side at any time, assuming the nonce count being 2 is not considered an adequate solution to this problem.

juha-h:
Richard Aas writes:

> The SIP auth code relies on the nonce parameter to distinguish between
> two situations that both results in a 401/407 response; wrong
> credentials and stale nonce. Only the later case should trigger an
> immediate retry of the original request, otherwise it will result in a
> response/request loop (capped at 16 retries). As the credentials can
> change on the server side at any time, assuming the nonce count being
> 2 is not considered an adequate solution to this problem.

I agree that if the credentials change on the server side, then this
solution will result in re-tries.  However, credentials do not normally
change very often.  So there is trade-off: 16 retries in rarely
occurring situation or auth failure in common situation.

The only fool proof solution is the one that I suggested earlier:  make
it configurable if nonce reuse is tried at all.

juha-h:
I knew that this pull request will not be accepted, but based on comment. gave it a try anyway.

premultiply:
Oh no....!

fwjavox:
I had the same problem and found a very similar workaround but never created a merge request because it's not baresip but the server's fault. In our case this happens with Oracle SBCs. Waiting for Oracle to fix this is not an option for us.
We took the trade-off to not allow password changes at any time.

robert-scheck:
I'm also seeing this issue e.g. with AVM FRITZ!Box or Starface PBX (Asterisk based) and I wonder if there is a chance for reopening this PR, but adding a further option to `auth_handler()` which allows to enable the workaround if needed. This could be then at least used by API users (e.g. baresip) to enable the workaround using a configuration option, if the user is willing to do the trade-off between 16 retries vs. auth failures.

PR #229, opened 2019-10-08
Rfc3581 received rport parsing

wnetbal:
Parse the received and rport parameter in the registration answer defined in RFC3581 from the server to use it in a later commit for the contact header.


alfredh:
Many thanks for your patch

critical defects.

PR #230, opened 2019-10-08
sipsess: add handler for contact

wnetbal:
Adding handler sipsess_contact_h that returns a
struct sip_contact used by sip_contact_print().
The contact handler is used contact information
saved in an application using re.


alfredh:
Many thanks for your patch

critical defects.

PR #235, opened 2020-01-24
WIP: Add support for in-call NOTIFY's

gaaf:
Tried to get some feedback on the mailinglist, but got no response. Trying here instead.
If the PR is acceptable as-is, please let me know, I'll cook up a better PR message.

I'm in the process of trying to support some PBX extensions to allow [remote 
control of the UA](https://github.com/alfredh/baresip/issues/758) by in-call NOTIFY's. The application to be controlled 
(baresip) is the one using re.

Unfortunately I could not find any documentation on architecture and/or call-
flows, and as a newcomer to re's code, I'm just guessing on where to add the 
needed functionality.

I would need to receive NOTIFY's that are part of the INVITE dialog, not a 
separate SUBSCRIBE dialog nor out-of-dialog. Judging by the function names I 
guessed I would need to intercept is somewhere in sipsess, just like in-dialog 
INFO messages. So I came up with the first patch in this PR.

Then I found out that baresip is also supporting REFER. Which, by the 
standard, is also using in-dialog NOTIFY's but these seem to be caught from 
within sipevent. Sipevent, by first glance, doesn't seem to have a relation to 
the sipsess of the current call and messaged are somehow coupled later on. 
Other uses of sipevent seem to suggest it is meant to be used for SUBSCRIBE 
dialogs, not INVITE based.

But due to the handling in sipevent instead of sipsess of the REFER's NOTIFY's 
I had to come up with the 2nd patch to let this still work. This is what gave rise to 
doubt about the chosen direction and about the architecture of re in general.

Note that my code is only interested in NOTIFY's which are in-dialog to the 
INVITE dialog. Out-of-dialog or NOTIFY's belonging to another call/dialog are 
not of interest and would preferably not even reach the code.

Are there separate code paths for in- and out-of-dialog messages? Or is it one 
big pile of messages and is application code supposed to find out what belongs 
where?

What would be the right place to intercept only(!) in-dialog messages 
belonging to a dialog started by an INVITE? And with "right" I mean such that 
it would be acceptable upstream, as the code is intended to be upstreamed when 
finished.

### DIFF
--- a/src/sipsess/listen.c
+++ b/src/sipsess/listen.c
@@ -245,7 +245,11 @@ static void invite_handler(struct sipsess_sock *sock,
 }
 
 
-static void notify_handler(struct sipsess_sock *sock, const struct sip_msg *msg)
+/* Handle NOTIFY message
+ * Return false if message could not be handled, it will then fall through to
+ * out-of-dialog message handling.
+ */
+static bool notify_handler(struct sipsess_sock *sock, const struct sip_msg *msg)
 {
 	struct sip *sip = sock->sip;
 	struct sipsess *sess;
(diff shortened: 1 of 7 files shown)


src/sipsess/listen.c
```diff
@@ -245,6 +245,34 @@ static void invite_handler(struct sipsess_sock *sock,
...
+static bool notify_handler(struct sipsess_sock *sock, const struct sip_msg *msg)
+{
+	struct sip *sip = sock->sip;
+	struct sipsess *sess;
+
+	sess = sipsess_find(sock, msg);
+	if (!sess || sess->terminated) {
+		(void)sip_reply(sip, msg, 481, "Call Does Not Exist");
+		return true;
+	}
```
alfredh: perhaps return false here, so that the NOTIFY message can be handled
by the next listener ?

gaaf: done

alfredh:
Many thanks for your patch

critical defects.

PR #238, opened 2020-02-24
ice: prevent invalid state selection

sreimers:


alfredh:
I am planning to clean up the ICE code at some point.
lets fix it all then.

PR #242, opened 2020-03-24
Win32 fixes

njeisecke:
Hi Alfred,

these are my fixes to make libre compile with VS 2017. It also disables memory tracking to avoid crashes due to missing mutex. Not sure if other parts would need a mutex too.

### DIFF
--- a/src/jbuf/jbuf.c
+++ b/src/jbuf/jbuf.c
@@ -89,8 +89,10 @@ static void frame_alloc(struct jbuf *jb, struct frame **f)
 		f0 = le->data;
 
 		STAT_INC(n_overflow);
+#if JBUF_STAT
 		DEBUG_INFO("drop 1 old frame seq=%u (total dropped %u)\n",
 			   f0->hdr.seq, jb->stat.n_overflow);
+#endif
 
 		f0->mem = mem_deref(f0->mem);
 		list_unlink(le);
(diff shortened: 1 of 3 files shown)


Encamy:
Ty for updating project!
Instead of increasing FD_SETSIZE to 2048 i whould like to recommend use patch https://github.com/creytiv/re/commit/49834b7afd0fe31660108271ef7657a34caa6a6d
I can create pull request with it
@alfredh what about having two `main.c`? One for all platforms and one for windows
This patch was tested and it is working stably

njeisecke:
@Encamy I have not tried that patch yet, but I've read some comments about the fd lookup being too slow.

Encamy:
@njeisecke 
I didn't notice slowness of fd lookup. For example, it is able to make 200 concurrent call without any difficulties. It is stable and reliable

src/jbuf/jbuf.c
```diff
@@ -89,8 +89,10 @@ static void frame_alloc(struct jbuf *jb, struct frame **f)
 		f0 = le->data;
 
 		STAT_INC(n_overflow);
+#if JBUF_STAT
```
alfredh: please remove this hunk

njeisecke: jbuf::stat is only available with JBUF_STAT defined (via RELEASE=1).

njeisecke: @alfredh Could you please have another look? Thanks!

PR #253, opened 2020-08-15
Sipreg expires

cspiel1:
This are a series of patches that belong to registration, de-registration and to re-registration.


cspiel1:
This are a series of patches that belong to registration, de-registration, re-registration interval.

cspiel1:
Maybe should go first to baresip/re.

PR #254, opened 2020-09-10
Fix for the fact that windows file descriptors do not follow possix numbering standard

johnjuuljensen:
Hey

So I tried to build and run under windows, but calling kept failing in fd_listen due to out of bounds check on the file descriptor.
I was pointed to this change: https://github.com/creytiv/re/commit/49834b7afd0fe31660108271ef7657a34caa6a6d
But thought it was somewhat invasive.

I ended up making this change instead. It's pretty simple and consists mostly of renaming of variables to clearly distinguish between what is an index into fhs and what is an actual descriptor handle.

The only new thing is the windows-only lookup_fd_index which maps file descriptors to indices.

Baresip compiles and runs and makes calls, when build with this version on windows SDK 10.

I have not tested anything under linux, not even compilation.

If you're interested, I can make any recommended changes and test it under linux.

Cheers,
John

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,19 +8,67 @@ compiler:
     - clang
     - gcc
 
+jobs:                                                                                                                                                                                         
+  include:
+      - os: linux
+        stage: ccheck
+        addons:
+          apt:
+            update: false
+        script:
+          - wget "https://raw.githubusercontent.com/baresip/baresip/master/test/ccheck.py"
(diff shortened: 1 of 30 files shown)


sreimers:
I like this approach, can you make a pull request on this repo:

https://github.com/baresip/re

johnjuuljensen:
Resubmitting on baresip/re

PR #259, opened 2021-06-07
add loongarch little endian

loongson-zjl:
add loongarch little endian


xen0n (changes requested):
This change is not needed, because the `#include <ctype.h>` already provides the correct `__BYTE_ORDER` for use.

PR #262, opened 2022-09-06
Add hash_remove and use in transp to fix #261

### LINKED ISSUE #261: SIGABRT after tcp_close_handler derefs conn one too many times (race condition on socket close)
I have observed the following issue in `re`:
1. Send a SIP message and store a reference to it.
   - Note: `conn_send` allocates a `struct sip_conn` with `nrefs=1` (from the allocation, zero refs from adding it to `ht_conn`).  Then the `struct sip_msg` stores another reference so `nrefs=2`.
2. Dereference the sip struct (`sip_close`).
3. Block main thread for a couple of seconds until remote party closes the socket cleanly.
  - This can occur when debug logging is enabled with a slow logging callback.
  - In the meantime, the remote party closes the socket, `tcp_close_handler` dereferences the connection (now `nrefs=1`) and `sip_close` flushes `ht_conn` which dereferences it again (now `nrefs=0` and the connection is freed).
4. Dereference the SIP message
  - but `struct sip_conn` has already been freed. 

Result: SIGABRT after:
```
mem: mem_deref: magic check failed 0xb5b5b5b5 (0x120d5c0)
```

Reason: the connection struct has been dereferenced once by `tcp_close_handler` and once too many by `sip_close` flushing `sip->ht_conn`.

I can fix this by removing [the call to `mem_deref` from `tcp_close_handler` in `transp.c`](https://github.com/creytiv/re/blob/cae2bcb6c03499397c0d2eb7a906b4d4890feb15/src/sip/transp.c#L495).  But, is that a good change?  Could it cause memory leaks?

balgillo:
Fix for https://github.com/creytiv/re/issues/261

- Add a new `hash_remove` function which works like `hash_flush` but for a single hashtable entry.
- Remove `mem_deref` calls from `transp.c` and use `hash_remove` in `close_conn` instead.
  - This ensures that when the `conn` has already been removed from the hashtable, e.g. by `hash_flush(sip->ht_conn)`,
    there is no extra `mem_deref` in `transp.c`.

# Testing done
- Compiles fine
- `retest -r` passes
- Tested within two proprietary, internal test suites written in Python
- I have not reproduced this issue in pure C code due to the lack of `request` tests in `retest`.

### DIFF
--- a/include/re_hash.h
+++ b/include/re_hash.h
@@ -12,6 +12,7 @@ struct pl;
 int  hash_alloc(struct hash **hp, uint32_t bsize);
 void hash_append(struct hash *h, uint32_t key, struct le *le, void *data);
 void hash_unlink(struct le *le);
+void hash_remove(struct le *le);
 struct le *hash_lookup(const struct hash *h, uint32_t key, list_apply_h *ah,
 		       void *arg);
 struct le *hash_apply(const struct hash *h, list_apply_h *ah, void *arg);
--- a/src/hash/hash.c
+++ b/src/hash/hash.c
@@ -96,6 +96,23 @@ void hash_unlink(struct le *le)
 
 
 /**
+ * If the element is present in the hashmap table, unlink it and deref the object,
+ * otherwise do nothing.  This is like hash_flush but for a single entry.
+ *
+ * @param le     List element
+ */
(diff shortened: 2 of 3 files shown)


Orkhan647:
👍

